Issue #61, opened 2020-05-11
Undocked plugins have transparent background


Sarath18:
Plugins in the main window, when undocked, have a transparent background which makes them less readable if they in front of some other plugin.
There are two possible implementations in my mind to solve them:
1. Adding a color palette on the floating plugin toolbar. User can select the background color along with the transparency.
2. Adding color palette with the transparency in plugin settings (right click -> settings).

The color palette option in the toolbar is easily accessible and noticeable but on the other hand color palette placed in the settings menu seems like the right way to do. I would like inputs on this issue to go ahead with the actual implementation.

chapulina:
+1 to color palette on the plugin settings menu, this is something that has come to mind before. Thanks!

PR #62, opened 2020-05-12
Add color palette to plugin settings to change background color

Sarath18:

Reference to issue #61 

### Description
Add color palette to select background color and transparency to the plugin settings. User can change the background in both docked and undocked state.

split

### Basic Info
Tested on: Ubuntu 18.04

### DIFF
--- a/include/ignition/gui/qml/IgnCard.qml
+++ b/include/ignition/gui/qml/IgnCard.qml
@@ -75,9 +75,9 @@ Pane {
   property string closeIcon: "\u2715"
 
   /**
-   * The plugin backgroung color
+   * The plugin backgroung color. Default: transparent
    */
-  property string cardBackground: "#ffffff"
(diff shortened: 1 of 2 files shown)


Sarath18:
Okay! Thanks. I'll look into it.

include/ignition/gui/qml/IgnCard.qml
```diff
@@ -74,6 +74,11 @@ Pane {
    */
   property string closeIcon: "\u2715"
 
+  /**
+   * The plugin backgroung color
+   */
+  property string cardBackground: "#ffffff"
```
chapulina, line 80: Can we keep the default value as `transparent`? Otherwise with this change `WorldStats` and `WorldControl` look like this:

whitebg

Sarath18: Okay. I'll do that.

PR #64, opened 2020-05-19
Move each plugin to its own directory

chapulina:
Builds on top of #63 

---

We were not installing plugin headers, so there's no need for them to be in the `include` folder.

This pull request also moves each plugin into its own directory so the code is more organized and we can more cleanly add plugins with multiple files.

The new `ign_gui_add_plugin` function helps reduce the boilerplate for each plugin.

Other changes:

* Removed QWT code that was commented out - we won't need that with QtQuick.
* Removed the `Requester` and `Responder` cc files since those plugins were not being compiled.

CC @claireyywang and @AmrElsersy


chapulina:
>  but seems like the CIs are failing

Those error messages are expected, because those tests are making sure that we properly handle invalid plugins and don't crash. I see some builds list these lines as "problems" though, so I ticketed an issue for that: https://github.com/ignition-tooling/release-tools/issues/205

claireyywang:
> > but seems like the CIs are failing
> 
> Those error messages are expected, because those tests are making sure that we properly handle invalid plugins and don't crash. I see some builds list these lines as "problems" though, so I ticketed an issue for that: [ignition-tooling/release-tools#205](https://github.com/ignition-tooling/release-tools/issues/205)

I see. I was able to build locally with no error.

PR #65, opened 2020-05-19
Re-enable publisher test

chapulina:
While working on #64 I noticed that the publisher test hadn't been ported from QWidgets to QtQuick. So I went ahead and ported it.

There are more tests in the same situation.


chapulina:
> other than the test failures on osx and windows.

Ah good call, this is only adding to the list of failing tests. I'll see if I can improve that.

chapulina:
> codecheck warning.

Argh, did you get these with `make codecheck`? There's a difference between that and `./tools/code_check.sh`, see https://github.com/ignition-tooling/release-tools/issues/148. The Ubuntu CI is warning-free, this should be our goal for now. We can't easily satisfy both checkers yet :confounded:

claireyywang:
>  There's a difference between that and `./tools/code_check.sh`, see [ignition-tooling/release-tools#148](https://github.com/ignition-tooling/release-tools/issues/148).

:laughing:  I always thought they were the same. Good to know!

chapulina:
@osrf-jenkins run tests por favor

chapulina:
@osrf-jenkins run tests

PR #69, opened 2020-06-01
Transport Topics Plugin

AmrElsersy:
Signed-off-by: AmrElsersy <amrelsersay@gmail.com>
Initial code for Transport Topics Plugin which lists all ign-transport topics currently published and their fields.
it also handles storing the field's data (this will be used by the main Plotting plugin)

Screenshot from 2020-06-01 04-14-43
@chapulina @claireyywang


src/plugins/transport_topics/TransportTopics.qml
```diff
@@ -0,0 +1,49 @@
...
+            text: model.name
+            color: "black"
+        }
+    }
+
+
+    verticalScrollBarPolicy: Qt.ScrollBarAlwaysOff
+    horizontalScrollBarPolicy: Qt.ScrollBarAlwaysOff
+    headerVisible: true
+    onDoubleClicked:  { tree.subscribeTopic(); tree.expandCollapseFolder(); }
```
chapulina, line 33: I'm wondering if we really want this plugin to subscribe to topics. In the end, the plugin that will need the data will be the plotting plugin.

So maybe all this plugin should do is displaying all the topics / fields available, and making them draggable to any other plugins. Then the other plugin can decide what to do with it. If it needs the data, it will subscribe to it. Makes sense? This is how the old [TopicViewer](https://github.com/ignitionrobotics/ign-gui/blob/ign-gui0/include/ignition/gui/plugins/TopicViewer.hh) worked.

That plugin also displayed [this icon](https://github.com/ignitionrobotics/ign-gui/blob/ign-gui0/include/ignition/gui/images/graph_line.svg) next to each plottable field. Feel free to copy it here if you want :wink:

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,301 @@
...
+TopicViewer::~TopicViewer()
+{
+}
+
+void TopicViewer::LoadConfig(const tinyxml2::XMLElement *)
+{
+  if (this->title.empty())
+    this->title = "Topic Viewer";
+}
+
```
claireyywang, line 84:
```suggestion

//////////////////////////////////////////////////
```
style nitpick: add line segregation between each function

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,301 @@
...
+//////////////////////////////////////////////////
+void TopicViewer::AddTopic(std::string _topic, std::string _msg)
+{
+  QStandardItem *topicItem = this->FactoryItem(_topic, "Topic");
+  QStandardItem *parent = this->dataPtr->model->invisibleRootItem();
+  parent->appendRow(topicItem);
+
+  // remove 'ignition.msgs.' from msg name
+  try {
```
claireyywang, line 120:
```suggestion
  try
  {
```
style guide #3: braces need to be on their own lines

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,301 @@
...
+/////////////////////////////////////////////////
+bool TopicViewer::IsPlotable(
+        const google::protobuf::FieldDescriptor::Type &_type)
+{
+  for (unsigned int j =0 ; j < this->plotableTypes.size() ; j++)
+  {
+    if (_type == this->plotableTypes[j])
+    {
+      return true;
+    }
```
claireyywang, line 266:
```suggestion
      return true;
```
nitpick

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,301 @@
...
+      msgFieldItem->setData(QVariant(true), PLOT_ROLE);
+    }
+  }
+}
+
+//////////////////////////////////////////////////
+QStandardItem* TopicViewer::FactoryItem(std::string _name,
+                                          std::string _type,
+                                          std::string _path,
+                                          std::string _topic)
```
claireyywang, line 180: nitpick: misalignment? same with some function params below

src/plugins/transport_topics/TransportTopics.hh
```diff
@@ -0,0 +1,91 @@
...
+{
+namespace plugins
+{
+  class TransportTopicsPrivate;
+
+  /// \brief handle Transport Topics Subscribing
+  class TransportTopics : public ignition::gui::Plugin
+  {
+    Q_OBJECT
+    private: std:: unique_ptr<TransportTopicsPrivate> data_ptr;
```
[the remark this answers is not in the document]
chapulina: Thanks for changing the case. Now we just need to move it to the bottom. The order should be:

* functions before variables
* public before protected before privat

Usually `dataPtr` is the very last one

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,128 @@
...
+                onClicked: mouse.accepted = false
+            }
+            */
+        }
+
+        ToolTip {
+            id: tool_tip
+            delay: 1000
+            text: (model === null) ? "Type ?" : "Type: " + model.type;
+            // visible: textArea.containsMouse
```
chapulina, line 107: It would be nice to enable these tooltips

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // =========== Delegates ============
+    rowDelegate: Rectangle {
+        color: (styleData.selected)? highlightColor : (styleData.row % 2 == 0) ? evenColor : oddColor
+        height: itemHeight;
+    }
+
+    itemDelegate: Item {
+        id :item
```
chapulina, line 69:
```suggestion
        id: item
```

AmrElsersy:
@chapulina I solved all style issues you mentioned 
and made the items draggable and the tooltip 

Screenshot from 2020-06-07 02-32-20

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,190 @@
...
+        id: item
+
+        // for fixing the item position
+        // item pos changes randomly when drag happens (with the copy drag)
+        anchors.top: parent.top
+        anchors.right: parent.right
+
+
+        Drag.dragType: Drag.Automatic
+        Drag.mimeData: { "text/plain" : "TopicViewer" }
```
chapulina, line 85: The dragging is working for me, but it always drags the text `TopicViewer`. I think it should drag information about the field being dragged so that the widget where it is dropped knows what topic to subscribe to and which field to plot.

Gazebo-classic uses a URI scheme like this: `topic?p=/fieldA/fieldB`

ploturi

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+    public: TopicsModel *model;
+  };
+}
+}
+}
+using namespace ignition;
+using namespace gui;
+using namespace plugins;
+
+TopicViewer::TopicViewer() :Plugin(), dataPtr(new TopicViewerPrivate)
```
chapulina, line 68:
```suggestion
TopicViewer::TopicViewer() : Plugin(), dataPtr(new TopicViewerPrivate)
```

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,153 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include <QStandardItem>
+#include <QStandardItemModel>
+#include <QHash>
+#include <QByteArray>
```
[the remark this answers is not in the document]
chapulina: Did you get a change to triage these, @AmrElsersy ?

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,148 @@
...
+#include <string>
+#include <vector>
+#include <deque>
+#include <memory>
+
+#include <ignition/gui/Application.hh>
+#include <ignition/gui/Plugin.hh>
+
+#include <ignition/msgs.hh>
+#include <ignition/transport.hh>
```
chapulina, line 32: It's preferable to never use these catch-all headers because they bring a lot of things that we won't use. Instead, we should just include the headers we need.

For example: `#include <ignition/msgs/Factory.hh>`

So let's remove all of these:

```
#include <ignition/msgs.hh>
#include <ignition/transport.hh>
#include <ignition/common.hh>
#include <ignition/gui.hh>
```

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,148 @@
...
+    public: void LoadConfig(const tinyxml2::XMLElement *) override;
+
+    /// \brief Create the fields model
+    private: void CreateModel();
+
+    /// \brief add a topic to the model
+    /// \param[in] _topic topic name to be displayed
+    /// \param[in] _msg topic's msg type
+    private: void AddTopic(const std::string &_topic,
+                           const std::string &_msg);
```
chapulina, line 72: Can we move all these private functions and variables to `TopicViewerPrivate`?

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,148 @@
...
+    /// \param[in] _msgName the displayed name of the field/msg
+    /// \param[in] _msgType field/msg type
+    private: void AddField(QStandardItem *_parentItem,
+                           const std::string &_msgName,
+                           const std::string &_msgType);
+
+    /// \brief factory method for creating an item
+    /// \param[in] _name the display name
+    /// \param[in] _type type of the field of the item
+    /// \param[in] _path a set of concatinate strings of parent msgs
```
chapulina, line 85:
```suggestion
    /// \param[in] _path a set of concatenate strings of parent msgs
```

It would be helpful to add an example here on the documentation

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,148 @@
...
+    private: void SetItemTopic(QStandardItem *_item);
+
+    /// \brief set the path/ID of the givin item starting from
+    /// the most topic parent to the field itself
+    /// \param[in] _item item ref to set its path
+    private: void SetItemPath(QStandardItem *_item);
+
+    /// \brief get the topic name of selected item
+    /// \param[in] _item ref to the item to get its parent topic
+    private: std::string TopicName(QStandardItem *_item);
```
chapulina, line 106: Can this function be `const`? The same for several ones below.

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,192 @@
...
+    horizontalScrollBarPolicy: Qt.ScrollBarAlwaysOff
+    headerVisible: false
+    headerDelegate: Rectangle {
+        visible: false
+    }
+    backgroundVisible: false;
+    TableViewColumn
+    {
+        role: "name";
+//        width: parent.width;
```
chapulina, line 57: Can this comment be removed?

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,192 @@
...
+
+    itemDelegate: Item {
+        id: item
+
+        // for fixing the item position
+        // item pos changes randomly when drag happens (with the copy drag)
+        anchors.top: parent.top
+        anchors.right: parent.right
+
+        Drag.mimeData: { "text/plain" :  model.topic + "," + model.path }
```
chapulina, line 83: How about using the same pattern as Gazebo-classic? `topic?p=/fieldA/fieldB`

That works like a URI, so it should be easier to break apart using the `ignition::common::URI` class.

chapulina: Also, I get messages like this when expanding items:

`[GUI] [Wrn] [Application.cc:649] [QT] file::/TopicViewer/TopicViewer.qml:83: TypeError: Cannot read property 'topic' of null`

It would be good to check if `model` is `null`.

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+  for (int i =0 ; i < msgDescriptor->field_count(); ++i)
+  {
+    auto msgField = msgDescriptor->field(i);
+
+    auto messageType = msgField->message_type();
+    if (messageType)
+    {
+      AddField(msgItem, msgField->name(), messageType->name());
+      igndbg << "name: " << msgField->name() <<
+                ", type: " << messageType->name() << std::endl;
```
chapulina, line 192: Can we remove all these debug messages? It's a bit spammy right now.

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+  {
+    std::vector<ignition::transport::MessagePublisher> infoMsgs;
+    this->dataPtr->node.TopicInfo(topics[i], infoMsgs);
+    std::string msgType = infoMsgs[0].MsgTypeName();
+    this->AddTopic(topics[i], msgType);
+  }
+}
+
+//////////////////////////////////////////////////
+void TopicViewer::SetPlottingMode(bool _mode)
```
chapulina, line 117: Is the concept of a "plotting mode" still necessary?

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+bool TopicViewer::PlottingMode()
+{
+  return this->plottingMode;
+}
+
+//////////////////////////////////////////////////
+void TopicViewer::AddTopic(const std::string &_topic,
+                           const std::string &_msg)
+{
+  // remove 'ignition.msgs.' from msg name
```
chapulina, line 132: Is this really necessary? It's possible to publish messages that are not `ignition.msgs`, so it may be helpful for the user to see the full type name.

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+  }
+  else
+  {
+    msgItem = this->FactoryItem(_msgName, _msgType);
+    _parentItem->appendRow(msgItem);
+  }
+
+  auto msg = ignition::msgs::Factory::New(_msgType);
+  if (!msg)
+      return;
```
chapulina, line 177: Do we want to print a warning if we can't create a message?

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+
+  auto msgDescriptor = msg->GetDescriptor();
+  if (!msgDescriptor)
+    return;
```
chapulina, line 181: This here probably should print and error

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+
+  for (int i =0 ; i < msgDescriptor->field_count(); ++i)
```
chapulina, line 183:
```suggestion
  for (int i = 0 ; i < msgDescriptor->field_count(); ++i)
```

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+    parent = parent->parent();
+  }
+
+  return _item->data(NAME_ROLE).toString().toStdString();
+}
+
+//////////////////////////////////////////////////
+std::string TopicViewer::ItemPath(QStandardItem *_item)
+{
+  std::deque<std::string> Path;
```
chapulina, line 266:
```suggestion
  std::deque<std::string> path;
```

:wink:

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+
+/////////////////////////////////////////////////
+bool TopicViewer::IsPlotable(
+        const google::protobuf::FieldDescriptor::Type &_type)
+{
+  for (unsigned int j =0 ; j < this->plotableTypes.size() ; ++j)
+  {
+    if (_type == this->plotableTypes[j])
+      return true;
+  }
```
chapulina, line 296: Just a suggestion, you could substitute the `for` loop with:

```
return std::find(this->plotableTypes.begin(), this->plotableTypes.end(), _type)
    != this->plotableTypes.end());
```

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,330 @@
...
+  return false;
+}
+
+//////////////////////////////////////////////////
+bool TopicViewer::IsPlotable(QModelIndex _index)
```
chapulina, line 301: Is this function ever used?

src/plugins/topic_viewer/TopicViewer.cc
AmrElsersy: not any more

src/plugins/topic_viewer/TopicViewer.cc
AmrElsersy: there is already a lot of msgs types that will be like that like [Map] for example or any msg that is not ign-msg

src/plugins/topic_viewer/TopicViewer.cc
AmrElsersy: not any more, i will remove it

src/plugins/topic_viewer/TopicViewer.hh
chapulina: I saw you made the input params const, which is great! But for the functions that are just getters, I think we could also make the function itself const, because they don't modify the object. For example:

`private: std::string TopicName(const QStandardItem *_item) const;`

AmrElsersy: done

src/plugins/publisher/CMakeLists.txt
```diff
@@ -4,6 +4,6 @@ ign_gui_add_plugin(Publisher
   QT_HEADERS
     Publisher.hh
   TEST_SOURCES
-    # Publisher_TEST.cc
+#    Publisher_TEST.cc
```
chapulina, line 7: Did you mean to edit the `Publisher` files?

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
+#include <gtest/gtest.h>
```
chapulina, line 1: Remember to add the copyright header to the test files too :wink:

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
...
+    plugins.clear();
+}
+
+TEST(TopicViewerTest, Model)
+{
+    // =========== Publish =================
+    transport::Node node;
+
+    // int
+    auto pubInt = node.Advertise<msgs::Int32>("/IntTopic");
```
chapulina, line 58: It's best practice to use `snake_case` for topic names, so I suggest:

```suggestion
    auto pubInt = node.Advertise<msgs::Int32>("/int_topic");
```

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
...
+            EXPECT_EQ(child->rowCount(), 2);
+
+            auto data = child->child(1);
+
+            EXPECT_EQ(data->data(NAME_ROLE), "data");
+            EXPECT_EQ(data->data(TYPE_ROLE), "int32");
+            EXPECT_EQ(data->data(PATH_ROLE), "data");
+            EXPECT_EQ(data->data(TOPIC_ROLE), "/IntTopic");
+            EXPECT_EQ(data->data(PLOT_ROLE), true);
+        }
```
chapulina, line 136: How about adding something like this, so we're sure no other topics are listed?

```
else
{
  EXPECT_TRUE(false);
}
```

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    auto root = model->invisibleRootItem();
+    ASSERT_NE(model, nullptr);
+
+    ASSERT_EQ(root->hasChildren(), true);
+
+    bool foundCollision = false;
+    bool foundInt = false;
+
+    for (int i = 0; i < root->rowCount(); ++i)
```
chapulina, line 101: You could add an expectation for the size of `rowCount` before this loop

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
...
+
+            EXPECT_EQ(data->data(NAME_ROLE), "data");
+            EXPECT_EQ(data->data(TYPE_ROLE), "int32");
+            EXPECT_EQ(data->data(PATH_ROLE), "data");
+            EXPECT_EQ(data->data(TOPIC_ROLE), "/IntTopic");
+            EXPECT_EQ(data->data(PLOT_ROLE), true);
+        }
+    }
+
+    EXPECT_EQ(foundCollision, true);
```
chapulina, line 139: You can use `EXPECT_TRUE` here:

```suggestion
    EXPECT_TRUE(foundCollision);
```

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    // int
+    auto pubInt = node.Advertise<msgs::Int32>("/IntTopic");
+    msgs::Int32 msgInt;
+
+    // collision
+    auto pub = node.Advertise<msgs::Collision> ("/CollisionTopic");
+    msgs::Collision msg;
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(300));
```
chapulina, line 65: Is this sleep needed here?

src/plugins/publisher/CMakeLists.txt
AmrElsersy: no no i was trying to run it to see something and i commented it again :D
i will modify that

src/plugins/topic_viewer/TopicViewer_TEST.cc
AmrElsersy: but the plugin list the topics that is currently in the transport 
so this will make the test case fails if we publishing anything to the transport when performing the testing
for example, if you run the test while you are opening ign-gazebo this will make it fails

chapulina: Ahh yes, that makes sense. One way to work around this is to set a different partition in the beginning of the test, so that all nodes starting during the test can only see other nodes also in the test. Could you give this a try for example?

`setenv("IGN_PARTITION", "ign-gazebo-test", 1);`

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,282 @@
...
+    this->dataPtr->node.TopicInfo(topics[i], infoMsgs);
+    std::string msgType = infoMsgs[0].MsgTypeName();
+    this->AddTopic(topics[i], msgType);
+  }
+}
+
+//////////////////////////////////////////////////
+QStandardItemModel *TopicViewer::Model()
+{
+    return reinterpret_cast<QStandardItemModel *>(this->dataPtr->model);
```
claireyywang, line 119:
```suggestion
  return reinterpret_cast<QStandardItemModel *>(this->dataPtr->model);
```
use two space indentation

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,282 @@
...
+    ignwarn << "Null Descriptor of Msg: " << _msgType << std::endl;
+    return;
+  }
+
+  for (int i = 0 ; i < msgDescriptor->field_count(); ++i)
+  {
+    auto msgField = msgDescriptor->field(i);
+
+    if (msgField->is_repeated())
+        continue;
```
claireyywang, line 170:
```suggestion
      continue;
```
use two space indentation

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,282 @@
...
+      auto msgFieldItem = this->FactoryItem(msgField->name(),
+                                            msgField->type_name());
+      msgItem->appendRow(msgFieldItem);
+
+      this->SetItemPath(msgFieldItem);
+      this->SetItemTopic(msgFieldItem);
+
+      // to make the plottable items draggable
+      if (this->IsPlotable(msgField->type()))
+          msgFieldItem->setData(QVariant(true), PLOT_ROLE);
```
claireyywang, line 188:
```suggestion
        msgFieldItem->setData(QVariant(true), PLOT_ROLE);
```
same as above

src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -0,0 +1,282 @@
...
+    pathString += path[path.size()-1];
+
+  return pathString;
+}
+
+/////////////////////////////////////////////////
+bool TopicViewer::IsPlotable(
+    const google::protobuf::FieldDescriptor::Type &_type)
+{
+    return std::find(this->plotableTypes.begin(), this->plotableTypes.end(),
```
claireyywang, line 275:
```suggestion
  return std::find(this->plotableTypes.begin(), this->plotableTypes.end(),
```

src/plugins/topic_viewer/TopicViewer.hh
```diff
@@ -0,0 +1,124 @@
...
+    /// \brief Destructor
+    public: ~TopicViewer();
+
+    /// \brief Documentaation inherited
+    public: void LoadConfig(const tinyxml2::XMLElement *) override;
+
+    /// \brief Create the fields model
+    private: void CreateModel();
+
+    public: QStandardItemModel *Model();
```
claireyywang, line 61: public before private? 

also would be nice to add `/// \brief` here

src/plugins/topic_viewer/TopicViewer_TEST.cc
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+
+    EXPECT_TRUE(foundCollision);
+    EXPECT_TRUE(foundInt);
+
+    // =========== Dynamic Adding / Removing =============
+
+    // Add
+    auto pubEcho = node.Advertise<msgs::StringMsg> ("/echo_topic");
```
chapulina, line 180:
```suggestion
    auto pubEcho = node.Advertise<msgs::Collision> ("/echo_topic");
```

This is currently failing to publish because of the type mismatch:

```
Node::Publisher::Publish() Type mismatch.
        * Type advertised: ignition.msgs.StringMsg
        * Type published: ignition.msgs.Collision
```

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,252 @@
...
+
+            font.pointSize: 12
+            anchors.leftMargin: 5
+            anchors.left: icon.right
+            y: icon.y
+        }
+
+        ToolTip {
+            id: tool_tip
+            delay: 1000
```
chapulina, line 172: The animation is quite slow right now :sweat_smile: How about making it much quicker:

```suggestion
            delay: 100
```

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,252 @@
...
+        else
+            tree.expand(index);
+    }
+
+    Transition {
+        id: expandTransition
+        NumberAnimation {
+            property: "y";
+            from: (tree.__listView.currentItem) ? tree.__listView.currentItem.y : 0;
+            duration: 1000;
```
chapulina, line 232:
```suggestion
            duration: 100;
```

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,252 @@
...
+            easing.type: Easing.OutQuad
+        }
+    }
+
+    Transition {
+        id: displacedTransition
+        NumberAnimation {
+            property: "y";
+            duration: 1000;
```
chapulina, line 241:
```suggestion
            duration: 100;
```

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,252 @@
...
+          height: itemHeight
+          width: itemHeight
+          color: "transparent"
+          Image {
+            id: branchImage
+            fillMode: Image.Pad
+            anchors.right: parent.right
+            anchors.verticalCenter: parent.verticalCenter
+            sourceSize.height: itemHeight * 0.5
+            sourceSize.width: itemHeight * 0.5
```
chapulina, line 196:
```suggestion
            sourceSize.width: itemHeight * 0.4
```

The icons on this plugin are bigger than the ones on the Entity Tree now. How about giving them the same size, multiplying by 0.4 instead? It will look like this:

iconsize_plot

src/plugins/topic_viewer/TopicViewer.qml
```diff
@@ -0,0 +1,252 @@
...
+        Image {
+            id: icon
+            source: "plottable_icon.svg"
+            height: itemHeight * 0.6
+            width: itemHeight * 0.6
+            y : itemHeight * 0.2
+            visible: (model === null) ? false : model.plottable
+        }
+
+        Text {
```
chapulina, line 157: The text is leaking when it's larger than the space available on the plugin. If you do this:

```.diff
diff --git a/src/plugins/topic_viewer/TopicViewer.qml b/src/plugins/topic_viewer/TopicViewer.qml
index 93aa675..7aba69e 100644
--- a/src/plugins/topic_viewer/TopicViewer.qml
+++ b/src/plugins/topic_viewer/TopicViewer.qml
@@ -164,12 +164,14 @@ TreeView {
             font.pointSize: 12
             anchors.leftMargin: 5
             anchors.left: icon.right
+            anchors.right: parent.right
             y: icon.y
+            elide: Text.ElideMiddle
         }
 
         ToolTip {
             id: tool_tip
```

Then the text will elide in the middle (with ...) when there isn't enough space:

elide_topic

cottsay:
This appears to have caused a build regression on Windows: https://build.osrfoundation.org/job/ign_gui-ci-win/58/consoleFull#-20872881438ff58640-3599-4406-a210-216932f1748c

FYI @AmrElsersy / @chapulina / @claireyywang

chapulina:
Thanks, @cottsay , compilation should be fixed on https://github.com/ignitionrobotics/ign-gui/pull/103 :crossed_fingers:

Issue #68, opened 2020-05-29
Port Render event from ign-gazebo


chapulina:
The [Render event](https://github.com/ignitionrobotics/ign-gazebo/blob/44951e3ddfd238f24182d4d80b1376f0d426bd43/include/ignition/gazebo/gui/GuiEvents.hh#L151-L161) was added to `ign-gazebo` so that various GUI plugins could run code that was guaranteed to run on `ign-rendering`'s main thread. This is important when making changes to the rendering scene.

This concept would be very helpful to have in `ign-gui` so that we can write pure `ign-gui` plugins that affect the rendering scene. This should allow us to port a few plugins that are currently in `ign-gazebo` but could be here, like the `GridConfig`.

---

The way the GUI events are working on `ign-gazebo` is a bit involved:

1. The `Scene3D` plugin [emits the Render event](https://github.com/ignitionrobotics/ign-gazebo/blob/44951e3ddfd238f24182d4d80b1376f0d426bd43/src/gui/plugins/scene3d/Scene3D.cc#L556-L558) from the rendering thread to the main window.
1. Other plugins can [filter all the main window events](https://github.com/ignitionrobotics/ign-gazebo/blob/44951e3ddfd238f24182d4d80b1376f0d426bd43/src/gui/plugins/grid_config/GridConfig.cc#L83-L84) and [do what they need](https://github.com/ignitionrobotics/ign-gazebo/blob/44951e3ddfd238f24182d4d80b1376f0d426bd43/src/gui/plugins/grid_config/GridConfig.cc#L87-L95) when they see a render event.

This is a bit involved and maybe even hacky, but it accomplishes cross-plugin communication without direct dependencies between them, on the same thread and without polluting the transport space.

PR #70, opened 2020-06-02
Add render event

Sarath18:
* Create new render event type

* Sends render event after each render call

Reference to issue #68

### DIFF
--- a/include/ignition/gui/GuiEvents.hh
+++ b/include/ignition/gui/GuiEvents.hh
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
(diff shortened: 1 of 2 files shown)


src/plugins/scene3d/Scene3D.cc
```diff
@@ -36,6 +36,10 @@
 #include <ignition/transport/Node.hh>
 
 #include "ignition/gui/Conversions.hh"
+#include "ignition/gui/Application.hh"
+#include "ignition/gui/MainWindow.hh"
+#include "ignition/gui/GuiEvents.hh"
```
chapulina, line 41: Nitpick: alphabetize

```
#include "ignition/gui/Application.hh"
#include "ignition/gui/Conversions.hh"
#include "ignition/gui/GuiEvents.hh"
#include "ignition/gui/MainWindow.hh"
```

include/ignition/gui/GuiEvents.hh
```diff
@@ -0,0 +1,47 @@
...
+      /// \brief Event called in the render thread of a 3D scene.
+      /// It's safe to make rendering calls in this event's callback.
+      class Render : public QEvent
+      {
+        public: Render()
+            : QEvent(kType)
+        {
+        }
+        /// \brief Unique type for this event.
+        static const QEvent::Type kType = QEvent::Type(QEvent::User);
```
chapulina, line 41: I just realized that we should make sure these types don't collide with the types on [ignition::gazebo::gui::GuiEvents](https://github.com/ignitionrobotics/ign-gazebo/blob/master/include/ignition/gazebo/gui/GuiEvents.hh#L76). 

Since `ign-gazebo` is counting up from `QEvent::User`, how about `ign-gui` starting from `QEvent::MaxUser` and counting down?

Sarath18: Okay. Should I add a comment in the code mentioning this?

chapulina: Yeah that's a good idea, thanks for adding the comment.

PR #80, opened 2020-06-12
[Bazel] Add Gui bazel build files

JShep1:
I have the tests commented out for now as there's a linking issue with Qt5 that I couldn't quite resolve locally.  It might have to do with needing to create a symbolic link or something similar.


BUILD.bazel
```diff
@@ -0,0 +1,337 @@
...
+    "//ign_bazel:qt.bzl",
+    "qt_resource",
+    "qt_cc_library",
+    "qt_cc_binary",
+)
+
+package(default_visibility = ["//visibility:public"])
+
+public_headers_no_gen = [
+    "//ign_gui/include/ignition/gui:Application.hh",
```
mjcarroll, line 27: This seems really strange, is the relative path not sufficient here?

BUILD.bazel
```diff
@@ -0,0 +1,337 @@
...
+)
+cc_library(
+    name = "ign_gui",
+    srcs = ["libignition-gui4.so"],
+    hdrs = public_headers,
+    includes = ["include"],
+    visibility = ["//visibility:public"],
+    linkopts = [
+        "-Wl,-soname,libignition4-gui.so",
+        "-ltinyxml2",
```
mjcarroll, line 276: I believe that you should be able to use `@tinyxml2` in the deps, rather than linkopts

BUILD.bazel
```diff
@@ -0,0 +1,337 @@
...
+        "@qt//:qt_qml",
+        "@qt//:qt_gui",
+        "@qt//:qt_opengl",
+    ],
+)
+[cc_test(
+    name = src.replace("/", "_").replace(".cc", "").replace("src_", ""),
+    srcs = [src],
+    linkopts = [
+        "-ltinyxml2",
```
mjcarroll, line 300: Same here.

include/ignition/gui/BUILD.bazel
```diff
@@ -0,0 +1,120 @@
...
+exports_files(["Enums.hh"])
+exports_files(["GuiEvents.hh"])
+exports_files(["Helpers.hh"])
+exports_files(["ign.hh"])
+exports_files(["MainWindow.hh"])
+exports_files(["Plugin.hh"])
+exports_files(["SearchModel.hh"])
+exports_files(["qt.h"])
+
+exports_files(["qtquickcontrols2.conf"])
```
mjcarroll, line 106: This can be a list, rather than multiple invocations, I believe.

src/Application.cc
```diff
@@ -306,24 +305,36 @@ bool Application::LoadPlugin(const std::string &_filename,
 
   common::SystemPaths systemPaths;
   systemPaths.SetPluginPathEnv(this->dataPtr->pluginPathEnv);
-
+  std::cout << "plugin path env" << this->dataPtr->pluginPathEnv << std::endl;
```
mjcarroll, line 308: Extra debug output can be deleted.

src/plugins/BUILD.bazel
```diff
@@ -0,0 +1,150 @@
...
+    "Publisher",
+    "Scene3D",
+    "TopicEcho",
+    "WorldControl",
+    "WorldStats",
+]
+
+[
+# NOTE the below genrule and following binary generations make the assumption that
+# all files within the plugin directory are titled %s.* where %s = plugin namei
```
mjcarroll, line 46:
```suggestion
# all files within the plugin directory are titled %s.* where %s = plugin name
``

BUILD.bazel
JShep1: Long story short, a build file must exist at the level of a qt resource `.qrc` file, since one exists in `include/ignition/gui`, the higher level build file sees the headers as being owned by the lower level build file, hence the out of the ordinary referencing.

BUILD.bazel
JShep1: That doesn't quite seem to be working for me

mjcarroll: Okay, we can skip this for now.

PR #81, opened 2020-06-15
KeyPublisher

mohamedsayed18:
a plugin that publish the key strokes of the keyboard with type Int32

### DIFF
--- a/src/plugins/key_publisher/KeyPublisher.cc
+++ b/src/plugins/key_publisher/KeyPublisher.cc
@@ -15,12 +15,10 @@
  *
 */
 
-#include <iostream>
-
+#include <ignition/msgs/int32.pb.h>
 #include <ignition/gui/Application.hh>
 #include <ignition/gui/MainWindow.hh>
 #include <ignition/plugin/Register.hh>
-#include <ignition/msgs/int32.pb.h>
 
 #include "KeyPublisher.hh"
 
--- a/src/plugins/key_publisher/KeyPublisher.cc
+++ b/src/plugins/key_publisher/KeyPublisher.cc
@@ -41,10 +41,10 @@ namespace gui
 
     /// \brief Publish keyboard strokes
     /// \param[in] key_press Pointer to the keyevent
-    public: void KeyPub(QKeyEvent *key_press)
+    public: void KeyPub(QKeyEvent *_keyPress)
     {
       ignition::msgs::Int32 Msg;
-      Msg.set_data(key_press->key());
+      Msg.set_data(_keyPress->key());
       pub.Publish(Msg);
     }
   };
--- a/src/plugins/key_publisher/KeyPublisher.cc
+++ b/src/plugins/key_publisher/KeyPublisher.cc
@@ -21,6 +21,7 @@
 #include <ignition/gui/MainWindow.hh>
 #include <ignition/plugin/Register.hh>
 #include <ignition/msgs/int32.pb.h>
+
 #include "KeyPublisher.hh"
 
 namespace ignition
@@ -34,7 +35,7 @@ namespace gui
 
     /// \brief Publisher
     public: ignition::transport::Node::Publisher pub;
-    
+
     /// \brief Topic
     public: std::string topic = "keyboard/keypress";
 
@@ -57,7 +58,8 @@ using namespace gui;
 KeyPublisher::KeyPublisher(): Plugin(), dataPtr(new KeyPublisherPrivate)
 {
   // Advertise publisher node
-  this->dataPtr->pub = this->dataPtr->node.Advertise<ignition::msgs::Int32>(this->dataPtr->topic);
+  this->dataPtr->pub = this->dataPtr->node.Advertise<ignition::msgs::Int32>
+    (this->dataPtr->topic);
 }
 
 /////////////////////////////////////////////////
--- a/src/plugins/key_publisher/KeyPublisher.cc
+++ b/src/plugins/key_publisher/KeyPublisher.cc
@@ -1,0 +1,60 @@
+#include <iostream>
+
+#include <ignition/gui/Application.hh>
+#include <ignition/gui/MainWindow.hh>
+#include <ignition/plugin/Register.hh>
+#include <ignition/msgs/int32.pb.h>
+
+#include "ignition/gui/plugins/KeyPublisher.hh"
+
+using namespace ignition;
+using namespace gui;
+//todo change the name of the class
+KeyPublisher::KeyPublisher(): Plugin()
+{
+  pub = node.Advertise<ignition::msgs::Int32>(topic);
+}
+
+
+KeyPublisher::~KeyPublisher()
+{}
+
+
+void KeyPublisher::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
+{
+  //it loads the XML file which contains the UI file of Qt
+  if (this->title.empty())
+  {
+    this->title = "Key tool";
+  }
+  ignition::gui::App()->findChild<ignition::gui::MainWindow *>()->QuickWindow()->installEventFilter(this);
+}
+
+
+
+bool KeyPublisher::eventFilter(QObject *_obj, QEvent *_event)
+{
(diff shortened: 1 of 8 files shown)


include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
chapulina, line 2:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

:wink:

include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef IGNITION_GUI_KEYPUBLISHER_HH_
```
chapulina, line 18: We've been using this pattern on all plugins:

```suggestion
#ifndef IGNITION_GUI_PLUGINS_KEYPUBLISHER_HH_
```

include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
...
+#include <ignition/gui/qt.h>
+#include <ignition/gui/Plugin.hh>
+#include <ignition/transport/Node.hh>
+
+
+namespace ignition
+{
+namespace gui
+{
+  class KeyPublisher : public ignition::gui::Plugin  //inherited from Qobject
```
chapulina, line 30: Could you document the class and all its functions? You can see examples on the other plugins.

include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
...
+    public: KeyPublisher();    //constructor
+    public: virtual ~KeyPublisher();   //destructor
+    
+    public: virtual void LoadConfig(const tinyxml2::XMLElement *_pluginElem) override;
+    protected: bool eventFilter(QObject *_obj, QEvent *_event) override;
+
+    //Publish message
+    public: ignition::transport::Node node;
+    public: std::string topic = "/gazebo/keyboard/keypress";
+    public: ignition::transport::Node::Publisher pub ;
```
chapulina, line 43: We've been using the PIMPL pattern for all plugins. It's helpful in case the headers are installed, which is not currently the case, but if we just follow the pattern it will be easier if we ever install it.

You can see examples on the other plugins. They have all private members inside a private class, i.e.

https://github.com/ignitionrobotics/ign-gui/blob/9da3dec08504774e7ebe9d446b0ab6acf38d299b/include/ignition/gui/plugins/Publisher.hh#L146-L148

Let me know if you have questions!

include/ignition/gui/plugins/KeyPublisher.qml
```diff
@@ -0,0 +1,9 @@
+import QtQuick 2.0
```
chapulina, line 1: Remember to add the copyright notice for QML files too

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
chapulina, line 2:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
...
+#include <ignition/gui/Application.hh>
+#include <ignition/gui/MainWindow.hh>
+#include <ignition/plugin/Register.hh>
+#include <ignition/msgs/int32.pb.h>
+#include "ignition/gui/plugins/KeyPublisher.hh"
+
+using namespace ignition;
+using namespace gui;
+
+KeyPublisher::KeyPublisher(): Plugin()
```
chapulina, line 29: Style nitpick: add `/////////////////////////////////////////////////` before every function

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
...
+{
+  pub = node.Advertise<ignition::msgs::Int32>(topic);
```
chapulina, line 31: Style nitpick: always use `this->` for member variables / functions. Once you use the PIMPL pattern, this will be `this->dataPtr->pub`.

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+
+KeyPublisher::~KeyPublisher()
+{
+}
+
+
+void KeyPublisher::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
```
chapulina, line 40: You can omit `_pluginElem` since it's not used. This prevents a warning from being printed:

```suggestion
void KeyPublisher::LoadConfig(const tinyxml2::XMLElement *)
```

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
...
+}
+
+
+
+bool KeyPublisher::eventFilter(QObject *_obj, QEvent *_event)
+{
+    if (_event->type() == QEvent::KeyPress)
+    {
+        QKeyEvent *keyEvent = static_cast<QKeyEvent*>(_event);
+        KeyPublisher::KeyPub(keyEvent);
```
chapulina, line 58: You can use `this`:

```suggestion
        this->KeyPub(keyEvent);
```

include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
...
+    public: void KeyPub(QKeyEvent *key_press);
```
chapulina, line 44: This function can be private, and be inside the `KeyPublisherPrivate` class once you create it :wink:

include/ignition/gui/plugins/KeyPublisher.hh
```diff
@@ -0,0 +1,49 @@
...
+
+    public: KeyPublisher();    //constructor
+    public: virtual ~KeyPublisher();   //destructor
+    
+    public: virtual void LoadConfig(const tinyxml2::XMLElement *_pluginElem) override;
+    protected: bool eventFilter(QObject *_obj, QEvent *_event) override;
+
+    //Publish message
+    public: ignition::transport::Node node;
+    public: std::string topic = "/gazebo/keyboard/keypress";
```
chapulina, line 42: Oh I know I suggested this originally, but now that the plugin is in `ign-gui`, it's weird to have `gazebo`. How about just:

```suggestion
    public: std::string topic = "/keyboard/keypress";
```

src/plugins/KeyPublisher.cc
```diff
@@ -0,0 +1,77 @@
...
+        return true;
```
chapulina, line 59: We shouldn't return here, otherwise this plugin "swallows" all the key presses on the entire UI. If we just let it continue and call `QObject::eventFilter` below, all other plugins have a chance to receive the key events too.

src/plugins/key_publisher/KeyPublisher.cc
```diff
@@ -0,0 +1,91 @@
...
+
+    /// \brief Publisher
+    public: ignition::transport::Node::Publisher pub;
+    
+    /// \brief Topic
+    public: std::string topic = "keyboard/keypress";
+
+    /// \brief Publish keyboard strokes
+    /// \param[in] key_press Pointer to the keyevent
+    public: void KeyPub(QKeyEvent *key_press)
```
chapulina, line 43: Style detail :wink: 

```suggestion
    public: void KeyPub(QKeyEvent *_keyPress)
```

Issue #74, opened 2020-06-10
Minimize plugins


chapulina:
As we start having larger plugins, they start taking more and more real estate on the window. For example, the plotting tool (#66) and the resource spawner (https://github.com/ignitionrobotics/ign-gazebo/issues/31) will be too large to keep on the left panel, and too large to be overlapped on top of the 3D scene when undocked.

Some plugins could be launched on their own window using 'ign gui -s'. This should work for plotting if all the user needs are topics. But plugins like the spawner will need to be on the same process as the 3D scene. So we need a way to make large plugins more manageable.

The idea on this ticket is to let plugins be collapsed / minimized. This applies to both docked and undocked plugins.

The minimize / maximize buttons could be on the plugin toolbar next to dock / undock.

Sarath18:
![minimize](https://user-images.githubusercontent.com/33201532/84362214-8200aa00-abea-11ea-8d9c-b0b200a78c31.gif)
Does this implementation meet a few of the ticket requirements?

I tried to add a button on the title bar to expand/collapse the plugin. I set the size of the plugin card rectangle to 50 (size of the title bar) to collapse the plugin. Is this correct? (feels kind of hacky way to do it)

chapulina:
Nice, that looks great! That sounds like a valid implementation to me. I believe it also works when the plugins are docked, right?

Looking at your gif, I also think it would make it super nice to apply a [NumberAnimation](https://doc.qt.io/qt-5/qml-qtquick-numberanimation.html) to the `height` when changing it, to make the transition smoother. See an example here:

https://github.com/ignitionrobotics/ign-gui/blob/75404e2ff46f4445a2f91b57f76033a0f6f6f2df/src/plugins/world_stats/WorldStats.qml#L123-L142

Sarath18:
> Nice, that looks great! That sounds like a valid implementation to me.

Great!

>  I believe it also works when the plugins are docked, right?

Not yet, will work on that.

> Looking at your gif, I also think it would make it super nice to apply a NumberAnimation to the height when changing it, to make the transition smoother. 

Yes, it is snappy at the moment. I'll try to add this feature as well.

PR #84, opened 2020-06-19
Minimize GUI Plugins

Sarath18:

Reference to issue #74

### DIFF
--- a/include/ignition/gui/qml/IgnCard.qml
+++ b/include/ignition/gui/qml/IgnCard.qml
@@ -355,6 +355,22 @@ Pane {
           to: 50
         }
       }
+    },
+    Transition {
+      from: "docked_collapsed"
+      to: "floating_collapsed"
+      SequentialAnimation {
+        ScriptAction {script: leaveDockedState()}
+        ScriptAction {script: enterFloatingState()}
+      }
+    },
+    Transition {
+      from: "floating_collapsed"
+      to: "docked_collapsed"
+      SequentialAnimation {
+        ScriptAction {script: leaveFloatingState()}
+        ScriptAction {script: enterDockedState()}
+      }
     }
   ]
 
@@ -367,8 +383,13 @@ Pane {
     var splitName = backgroundItem.addSplitItem();
     var splitItem = backgroundItem.childItems[splitName];
 
+    const collapsed = card.height === 50
+
     // Reparent to split
     card.parent = splitItem;
+
+    // Retain collapsed or expanded state
+    card.parent.Layout.minimumHeight = collapsed ? 50 : content.children[0].Layout.minimumHeight;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Sarath18:
The PR is currently in draft state. I am working on minimizing the gui plugins when in the docked state.

Sarath18:
Added two new states `d_collapsed` (docked collapsed) state and `f_collapsed` (floating collapsed) state and transition between these and the existing states. These states are working on config load as well.

min_plugin

include/ignition/gui/qml/IgnCard.qml
```diff
@@ -222,12 +242,20 @@ Pane {
...
     State {
       name: "docked"
     },
-
     State {
       name: "floating"
+    },
+    // Docked collapsed state
+    State {
+      name: "d_collapsed"
```
chapulina, line 254: How about being more verbose for clarity? `docked_collapsed` / `floating_collapsed` would be clearer to people not familiar with the code.

include/ignition/gui/qml/IgnSplit.qml
```diff
@@ -328,6 +328,8 @@ SplitView {
               {
                 Layout.minimumWidth = child.Layout.minimumWidth;
               }
+              // Set child height to minimu height
```
chapulina, line 331:
```suggestion
              // Set child height to minimum height
```

include/ignition/gui/qml/IgnCard.qml
```diff
@@ -39,6 +39,11 @@ Pane {
    */
   property bool showCloseButton: true
 
+  /**
+   * True to have a minimize button
+   */
+  property bool showMinimizeButton: true
```
chapulina, line 45: How about calling it `showCollapseButton`? Just so the API is consistent. Users are able to set this, as well as the state, from config files.

Sarath18:
> I noticed that the last docked plugin can't be minimized. Did you notice that? It looks like the internal state is collapsed, even though the icon doesn't change, because the height isn't changing.

Yes, you are right. I looked at it again and found that this case only happens when the sum of minimum heights of the plugins loaded (in split) is less than the height of the app window. If we have many expanded plugins before the last collapsed plugins, the expand and collapse work just fine. But the weird part is if you close the left split and try to expand/close the last (or any) plugins, it works as intended. Here is a gif describing the above-mentioned details.

min_bug

I am not sure why is this happening, can you help me find the issue?

chapulina:
> can you help me find the issue?

Sure, I'll take a look this week :wink:

Sarath18:
> Sure, I'll take a look this week :wink: 

Thank you for the help, but I believe I fixed it in the last commit 95670a7. Can you please review it?

include/ignition/gui/qml/IgnCard.qml
```diff
@@ -376,11 +485,59 @@ Pane {
         }
         visible: card.showDockButton && !card.standalone
         onClicked: {
-          const docked = card.state === "docked"
+          const docked = (card.state === "docked" || card.state === "docked_collapsed")
           card.state = docked ? "floating" : "docked"
```
chapulina, line 489: How do you feel about maintaining the collapsed / expanded state when docking / undocking?

For example, if it's currently `docked_collapsed` and I undock, I'd expect to remain collapsed, so I'd become `floating_collapsed`. The same goes when re-docking.

PR #85, opened 2020-06-22
Fixed left menu events

ahcorde:
In Ubuntu 20.04 the lef menu is not working. 

fail-gazebo

Probably because the keyword action is reserved. With this small fix the menu is working again.


ahcorde:
:facepalm: Fixed

PR #89, opened 2020-06-29
Plotting interface

AmrElsersy:
The Plotting Interface is responsible for Plotting the transport msgs that will be used from ign-gui and ign-gazebo plugins

it accepts the dragged fields from the TopicViewer plugin to register the fields to a specific chart

Screenshot from 2020-08-25 06-45-28

This is a UML Diagram for the code

Plot_ClassDiagram

@chapulina 
@claireyywang


include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,277 @@
...
+                spacing: 10
+            }
+        }
+
+        DropArea {
+            anchors.fill: parent
+            onDropped:
+            {
+                // topic and path is separated with ','
+                var topic_path = drop.text.split(",");
```
chapulina, line 52: We should do some checking to see if `drop.text` has a `,`. If it doesn't, print a warning and return.

src/plugins/plotting/GuiPlotting.cc
```diff
@@ -0,0 +1,36 @@
...
+
+GuiPlotting::~GuiPlotting()
+{
+}
+
+//////////////////////////////////////////
+GuiPlotting::GuiPlotting() : Plugin()
+{
+}
+
```
chapulina, line 32: Let's override the `LoadConfig` function so that you can set the title to something like  `Plotting`

src/plugins/plotting/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+ign_gui_add_plugin(GuiPlotting
```
chapulina, line 1: How about calling this plugin `TransportPlotting`? That may be clearer to end users.

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+  /// \brief get the registered topics
+  /// \return topics list
+  public: const std::map<std::string, Topic*> &Topics();
+
+  private: std::unique_ptr<TransportPrivate> dataPtr;
+};
+
+class PlottingIfacePrivate;
+
+class PlottingInterface : public QObject
```
chapulina, line 148: Please document this class. It's especially important to document it because it's being installed and any downstream user could use it.

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+{
+  Q_OBJECT
+
+  /// \brief current plotting time
+  float time;
+
+  /// \brief timer to update the plotting each time step
+  QTimer* timer;
```
chapulina, line 156: Let's move these member variables to `dataPtr` (i.e. `PlottingIfacePrivate`). This will allow us to change them even after releasing this.

include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,277 @@
+import QtQuick 2.0
```
chapulina, line 1: Missing copyright for the QML files :wink:

include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        // initial animation
+        Component.onCompleted:{
+            initialAnimation.start();
+        }
+        NumberAnimation {
+            id: initialAnimation
+            running: false
+            target: view
```
chapulina, line 218: When I add a new chart I get this warning:

`[GUI] [Wrn] [Application.cc:649] [QT] qrc:/qml/Chart.qml:218: ReferenceError: view is not defined`

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+  public: Q_INVOKABLE void setSimTime(double _timeout);
+
+  /// \brief plot a point to a chart
+  /// \param[in] _chart chart ID
+  /// \param[in] _fieldID field path ID
+  /// \param[in] _x x coordinates of the plot point
+  /// \param[in] _y y coordinates of the plot point
+  signals: void plot(int _chart, QString _fieldID, double _x, double _y);
+
+  private: std::unique_ptr<PlottingIfacePrivate> dataPtr;
```
chapulina, line 197:
```suggestion
  /// \brief Private data member.
  private: std::unique_ptr<PlottingIfacePrivate> dataPtr;
```

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+
+  /// \brief check the plotable types and get data from reflection
+  private: double PlotData(const google::protobuf::Message &_msg,
+                           const google::protobuf::FieldDescriptor *field);
+};
+
+class TransportPrivate;
+
+/// \brief handle Transport Topics Subscribing for one object(Chart)
+class Transport
```
chapulina, line 110: Can the `Field`, `Topic` and `Transport` classes be moved to another `.hh` file under the `src` folder? The least we have in this header the better, because everything under `include` in installed and can't be changed after it is released.

include/ignition/gui/resources.qrc
```diff
@@ -1,21 +1,21 @@
...
+        <file>qml/PlottingInterface.qml</file>
+        <file>qml/PluginMenu.qml</file>
+        <file>qml/SideDrawer.qml</file>
+        <file>qml/StandaloneDialog.qml</file>
+        <file>qml/StyleDialog.qml</file>
+        <file>qml/images/ignition_logo_50x50.png</file>
+        <file>qml/images/drawer.png</file>
+        <file>qml/images/menu.png</file>
+        <file>qml/Chart.qml</file>
+    </qresource>
```
chapulina, line 20: Are all these indentation changes necessary?

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+#include <bits/stdc++.h>
+#include <google/protobuf/message.h>
+#include <google/protobuf/descriptor.h>
+#include <ignition/transport/Node.hh>
+#include <ignition/transport/MessageInfo.hh>
+#include <ignition/transport/Publisher.hh>
+#include <map>
+#include <set>
+#include <string>
+#include <memory>
```
claireyywang, line 33: alphabetize, and need to follow order of headers https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes 

I'd recommend following this rule for all other headers in files to avoid codecheck warnings on CI.

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+  private: double value;
+
+  /// \brief Registered Charts to that field
+  private: std::set<int> charts;
+
+  /// \brief Constructor
+  public: Field();
+
+  /// \param[in] _value the set value
+  public: void SetValue(double _value);
```
claireyywang, line 51: missing `/// \brief`. maybe `const double _value`?

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+
+  /// \return value of the field
+  public: double Value();
```
claireyywang, line 54: recommend function name start with verb if possible, ex. `GetValue()`. missing `/// \brief`. also, can this be a `const` function?

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+
+  /// \brief Register a chart that plot that field
+  public: void AddChart(int _chart);
```
claireyywang, line 57: missing `/// \param[in]`

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+
+  /// \brief UnRegister a chart from plotting that field
+  /// \return the count of registered charts
+  public: void RemoveChart(int _chart);
```
claireyywang, line 61: `void` function doesn't need `/// \return`

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+{
+namespace gui
+{
+class Field
+{
+  /// \brief value of that field
+  private: double value;
+
+  /// \brief Registered Charts to that field
+  private: std::set<int> charts;
```
claireyywang, line 45: `public` -> `protected` -> `private` https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,182 @@
...
+    EXPECT_EQ(int(topics.size()), 1);
+}
+
+//// test class [ Under Progress ]
+//class PlotTest : public QObject
+//{
+//    Q_OBJECT
+//    public : PlotTest(PlottingInterface *plotIface)
+//    {
+//        connect(plotIface, SIGNAL(plot()), this, SLOT(plot()));
```
claireyywang, line 146: @chapulina during last week's meeting, @AmrElsersy asked about how to connect `SIGNAL` and `SLOT` in this function cuz this line doesn't work. I vaguely remember doing it many months ago and there was a tricky syntax involved https://doc.qt.io/qt-5/signalsandslots.html#signals-and-slots-with-default-arguments but am not sure if that is the cause of issue. Any suggestions?

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,203 @@
...
+
+  /// \brief UnRegister a chart from plotting that field
+  /// \return the count of registered charts
+  public: void RemoveChart(int _chart);
+
+  /// \brief size of registered charts
+  /// \return charts size
+  public: int ChartCount();
+
+  public: std::set<int> &Charts();
```
claireyywang, line 67: missing documentation on this function

include/ignition/gui/PlottingInterface.hh
AmrElsersy: ok but how can i include them after moving them to a .hh file in the /src folder ?

include/ignition/gui/resources.qrc
AmrElsersy: No :D 
i just add one line in the middle and all these changes are added
they don't have any changes except one line

include/ignition/gui/PlottingInterface.hh
chapulina: For example:

* `include/ignition/gui/PlottingInterface.hh`
    * Declares `PlottingInterface`
    * Doesn't need `Field`, `Topic` or `Transport`
* `src/PlottingTransport.hh`
   * Declares `Field`, `Topic` and `Transport` (just like you have here, but on a separate file)
* `src/PlottingTransport.cc`
    * Implements `Field`, `Topic` and `Transport`
    * Includes `#include "PlottingTransport.hh"`
* `src/PlottingInterface.cc`
    * Includes `#include "PlottingTransport.hh"`
    * Includes `#include "ignition/gui/PlottingInterface.hh"`
    * Implements `PlottingInterface`
    * Uses `Field`, `Topic` or `Transport`

include/ignition/gui/resources.qrc
chapulina: Does your editor have some setting to convert spaces into tabs or something like that? :monocle_face: I can help you revert these.

include/ignition/gui/PlottingInterface.hh
AmrElsersy: It was GetValue before but Louise recommended to be like "Value" instead :D

include/ignition/gui/PlottingInterface.hh
AmrElsersy: yeah, it was not void before, i just forgot to remove the comment :D

include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,547 @@
...
+    property int chartID: -1
+    property bool multiChartsMode: false
+    color: "transparent"
+    signal subscribe(real Id, string topic, string path);
+    signal unSubscribe(real Id, string topic, string path);
+    signal componentSubscribe(string entity, string typeId, string type, string attribute, real Id);
+    signal componentUnSubscribe(string entity, string typeId, string attribute, real Id);
+    signal clicked(real Id);
+
+    function appendPoint(_fieldID, _x, _y)
```
chapulina, line 35: Can you document all these functions and properties? Use the `/** */` notation like here:

https://github.com/ignitionrobotics/ign-gui/blob/58cedf4c96b4baf46a0c5c79234faf9e3dc4e924/src/plugins/world_control/WorldControl.qml#L50-L53

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,250 @@
...
+          chartObject.width = Qt.binding( function() {return chartsLayout.width});
+          // to change the height of each vertical chart
+          chartsLayout.heightFactor ++ ;
+      }
+
+      // Chart ID
+      chartObject.chartID = main.idIncrementor;
+      charts[idIncrementor] = chartObject;
+      // Signals and Slots
+      chartObject.subscribe.connect(main.onSubscribe);
```
chapulina, line 165: Just curious, you couldn't connect directly to `PlottingIface`'s `subscribe` slot here directly? The same goes for all the other connections being wrapped.

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+    auto fields = topic.second->Fields();
+
+    for (auto field : fields)
+    {
+      auto charts = field.second->Charts();
+
+      for (auto chart : charts)
+      {
+        QString fieldFullPath = QString::fromStdString(
+        topic.first + "-" + field.first);
```
claireyywang, line 451:
```suggestion
          topic.first + "-" + field.first);
```

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,250 @@
...
+      chart.parent = rowChartsLayout;
+  }
+
+  function onSubscribe(Id, topic, path)
+  {
+      PlottingIface.subscribe(topic, path, Id);
+  }
+  function onUnSubscribe(Id, topic, path)
+  {
+      PlottingIface.unsubscribe(topic, path, Id);
```
chapulina, line 192: If we really need to forward these calls, can we order the parameters the same way on both of them?

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,250 @@
...
+      charts[idIncrementor] = chartObject;
+      // Signals and Slots
+      chartObject.subscribe.connect(main.onSubscribe);
+      chartObject.unSubscribe.connect(main.onUnSubscribe);
+      chartObject.componentSubscribe.connect(main.onComponentSubscribe);
+      chartObject.componentUnSubscribe.connect(main.onComponentUnSubscribe);
+      chartObject.clicked.connect(main.onClicked);
+  }
+
+  function reallocateChart(chart)
```
chapulina, line 172: Style: prefix all input arguments, for all functions, with `_`:

```suggestion
  function reallocateChart(_chart)
```

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+bool Transport::TopicFound(const std::string &_topic) const
+{
+  // check if the topic exist
+  std::vector<std::string> topics;
+  this->dataPtr->node.TopicList(topics);
+  auto foundTopic = std::find(topics.begin(), topics.end(), _topic);
+  // topic is unsubscribed ... update the model
+  if (foundTopic == topics.end())
+  {
+    // TODO(Amr) : unsubscribe from the topic and remove its UI fields
```
claireyywang, line 337: Is it still a todo?

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+      if (valueMsg)
+        valueMsg = ref->MutableMessage
+                (const_cast<google::protobuf::Message *>(valueMsg), field);
+      else
+      {
+          valueMsg = ref->MutableMessage
+                  (const_cast<google::protobuf::Message *>(&_msg), field);
+      }
+
+        ref = valueMsg->GetReflection();
```
chapulina, line 216: oops, more bad indentation here and above

```suggestion
      ref = valueMsg->GetReflection();
```

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+{
+}
+
+////////////////////////////////////////////
+Transport::~Transport()
+{
+  // unsubscribe from all topics in the transport
+  for (auto topicIt = this->dataPtr->topics.begin();
+    topicIt != this->dataPtr->topics.end(); ++topicIt)
+    this->dataPtr->node.Unsubscribe(topicIt->first);
```
claireyywang, line 277:
```suggestion
      this->dataPtr->node.Unsubscribe(topicIt->first);
```

src/PlottingInterface.cc
chapulina, line 337: This sounds like something strange to do inside this function. As a user, I'd expect just to receive `true` / `false`, without any other side effects.

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+    {
+      std::string fieldName = fieldFullPath[i];
+
+      auto field = msgDescriptor->FindFieldByName(fieldName);
+
+      msgDescriptor = field->message_type();
+
+      if (valueMsg)
+        valueMsg = ref->MutableMessage
+                (const_cast<google::protobuf::Message *>(valueMsg), field);
```
claireyywang, line 209:
```suggestion
      if (valueMsg)
      {
        valueMsg = ref->MutableMessage
          (const_cast<google::protobuf::Message *>(valueMsg), field);
      }
```

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+      else
+      {
+          valueMsg = ref->MutableMessage
```
claireyywang, line 212:
```suggestion
        valueMsg = ref->MutableMessage
```

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+    this->dataPtr->name = _name;
+}
+
+//////////////////////////////////////////////////////
+Topic::~Topic()
+{
+}
+
+//////////////////////////////////////////////////////
+std::string Topic::Name() const
```
claireyywang, line 148:
```suggestion
std::string &Topic::Name() const
```

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,168 @@
...
+    vector3d->set_x(10);
+    vector3d->set_z(15);
+    pose->set_allocated_position(vector3d);
+    msg.set_allocated_pose(pose);
+
+    // publish to call the topic::Callback
+    pub.Publish(msg);
+
+    // wait for callback
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
chapulina, line 109: This test is a bit fragile and it will probably fail on slower machines. What we usually do it wait in a loop until a certain condition is reached, with a maximum timeout. For example:

https://github.com/ignitionrobotics/ign-gui/blob/58cedf4c96b4baf46a0c5c79234faf9e3dc4e924/src/plugins/publisher/Publisher_TEST.cc#L125-L132

Here, instead of `received`, you could check the value of a field.

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,168 @@
...
+    transport.Unsubscribe("/collision_topic", "pose-position-z", 1);
+    transport.Unsubscribe("/collision_topic", "pose-position-x", 1);
+
+    topics = transport.Topics();
+    EXPECT_EQ(static_cast<int>(topics.size()), 1);
+}
+
+TEST(PlottingInterfaceTest, PlottingIface)
+{
+//    auto PlotIface = new ignition::gui::PlottingInterface();
```
chapulina, line 144: What are you trying to test here? Anything we can help with?

src/PlottingInterface.cc
```diff
@@ -0,0 +1,466 @@
...
+    for (int i = 0; i < pathSize-1 ; i++)
+    {
+      std::string fieldName = fieldFullPath[i];
+
+      auto field = msgDescriptor->FindFieldByName(fieldName);
+
+      msgDescriptor = field->message_type();
+
+      if (valueMsg)
+        valueMsg = ref->MutableMessage
```
chapulina, line 208: Always wrap multiple lines in `{}`

src/PlottingInterface.cc
chapulina, line 216: Is there a chance we get here and `valueMsg` is `nullptr`? Maybe is `MutableMessage` returns null? In that case, we should check that the pointer is valid before using it.

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,168 @@
...
+
+    // test the deletion of the topic if it has no fields
+    transport.Unsubscribe("/collision_topic", "pose-position-z", 1);
+    transport.Unsubscribe("/collision_topic", "pose-position-x", 1);
+
+    topics = transport.Topics();
+    EXPECT_EQ(static_cast<int>(topics.size()), 1);
+}
+
+TEST(PlottingInterfaceTest, PlottingIface)
```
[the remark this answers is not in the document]
AmrElsersy: I think it could be removed

src/plugins/plotting/TransportPlotting.hh
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_GUI_PLUGINS_GuiPlotting_HH_
+#define IGNITION_GUI_PLUGINS_GuiPlotting_HH_
```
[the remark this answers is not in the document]
AmrElsersy: It is already

include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,547 @@
...
+    Rectangle{
+        id: infoRect
+        width: parent.width
+        height: (multiChartsMode) ? 0 : parent.height / 10
+        color: (Material.theme == Material.Light) ? Material.color(Material.Grey,Material.Shade200)
+                                                  : Material.color(Material.BlueGrey, Material.Shade800)
+
+        Text {
+            id: guideText
+            text: qsTr("Drag & Drop from TopicViewer | ComponentInspector")
```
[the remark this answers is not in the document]
AmrElsersy: Yeah but it should be in the same format that is processed by the plugin
which is  "topic,field"

but ok i updated it

include/ignition/gui/qml/PlottingInterface.qml
AmrElsersy: Yes i can do that, i do this to make the interactions between the UI and the Class through the PlottingInterface.qml only 👀

src/PlottingInterface_TEST.cc
AmrElsersy: that was an old problem, but it tests small functionality, and i think it can be removed

src/PlottingInterface.cc
AmrElsersy: No, it will get updated each iteration due to the "else" part
and this loop is iterate over the field path and it always have a path of non-zero size

but it is ok to check, i will do it

src/PlottingInterface.cc
AmrElsersy: it was an old function i updated it to unsubscribe from all the topics that are outdated

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,168 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include <gtest/gtest.h>
+#include <ignition/common/Console.hh>
+#include "test_config.h"
```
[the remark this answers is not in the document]
AmrElsersy: what do you mean ? :D

src/PlottingInterface_TEST.cc
```diff
@@ -0,0 +1,158 @@
...
+    vector3d->set_z(15);
+    pose->set_allocated_position(vector3d);
+    msg.set_allocated_pose(pose);
+
+
+    bool received = false;
+    std::function<void(const msgs::Collision &)> cb =
+        [&](const msgs::Collision &_msg)
+    {
+      EXPECT_EQ(_msg.pose().position().x(), 10);
```
claireyywang, line 109: got a compiler warning here on Ubuntu CI
```
/home/claire/gsoc_dome_ws/src/ign-gui/test/gtest/include/gtest/gtest.h:1444:11: warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
   if (lhs == rhs) {
```
I recommend to either cast the `x` value to `int`, or use `EXPECT_NEAR(_msg.pose().position().x(), 10, 1e-6)` to get rid of the warning

include/ignition/gui/PlottingInterface.hh
```diff
@@ -0,0 +1,259 @@
...
+  Q_OBJECT
+
+  /// \brief update the plotting each timeout of the timer
+  public slots: void UpdateGui();
+
+  /// \brief Constructor
+  public: explicit PlottingInterface();
+
+  /// \brief Destructor
+  public: ~PlottingInterface();
```
claireyywang, line 164:
```suggestion
  public: ~PlottingInterface() {};
```

src/PlottingInterface.cc
```diff
@@ -0,0 +1,482 @@
...
+  this->InitTimer();
+
+  App()->Engine()->rootContext()->setContextProperty("PlottingIface", this);
+}
+
+//////////////////////////////////////////////////////
+PlottingInterface::~PlottingInterface()
+{
+}
+
```
claireyywang, line 371:
```suggestion
```

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,546 @@
...
+                if (Object.keys(serieses).length === 0)
+                    continue;
+
+
+                // convert Serieses to Map of {series_name : points list}
+                // slots in cpp accepts QMap<QString, QVariant>
+                // QVariant could be a list instead of series object
+
+                var chartSerieses = {};
+                let seriesArray = [];
```
claireyywang, line 377: what's `let`, typo?

AmrElsersy: what is "typo" ? :D 
in this code i put the series's points into a list "seriesArray", as slots in cpp will not accept the series object itself

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,539 @@
...
+ * limitations under the License.
+ *
+*/
+import QtQuick 2.9
+import QtCharts 2.2
+import QtQuick.Controls 2.2
+import QtQuick.Controls.Styles 1.4
+import QtQuick.Controls.Material 2.1
+import QtQuick.Layouts 1.3
+import Qt.labs.platform 1.1
```
chapulina, line 23:
```suggestion
import Qt.labs.platform 1.0
```

Ubuntu Bionic has version 1.0.

I also had to install an extra dependency: `qml-module-qt-labs-platform`. Can you add it here?

https://github.com/ignitionrobotics/ign-gui/blob/58cedf4c96b4baf46a0c5c79234faf9e3dc4e924/.github/ci/packages.apt#L12-L22

include/ignition/gui/qml/PlottingInterface.qml
chapulina: I had to change the `let` to `var` for the plugin to load. It looks like Qt only suppors ES6 (which introduces `let`) from version 5.12, but on Ubuntu Bionic we're on Qt 5.9.5.

include/ignition/gui/qml/PlottingInterface.qml
AmrElsersy: ok done

include/ignition/gui/qml/Chart.qml
```diff
@@ -0,0 +1,670 @@
...
+            onPositionChanged: {
+                if (multiChartsMode)
+                    return
+
+                if (pressed)
+                {
+                    // percentage of amount of drag
+                    var xScroll = (mouseX - xHold) / (chart.plotArea.width)
+                    var yScroll = (mouseY - yHold) / (chart.plotArea.height)
+                    console.log(xScroll,yScroll)
```
chapulina, line 508: Let's remove this debug statement

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,538 @@
...
+
+      width: parent.width
+      anchors.topMargin: 10
+      anchors.top: rowCharts.bottom
+      anchors.bottom: parent.bottom
+      property int heightFactor: 0
+  }
+
+
+  Rectangle {
```
chapulina, line 76: Instead of using `Rectangle` + `Text` + `MouseArea`, consider using `ToolButton`, which has that all ready for you.

src/PlottingInterface.cc
```diff
@@ -0,0 +1,542 @@
...
+    for (auto const &field : fields)
+    {
+      auto charts = field.second->Charts();
+
+      for (auto const &chart : charts)
+      {
+        QString fieldFullPath = QString::fromStdString(
+          topic.first + "-" + field.first);
+        double x = this->dataPtr->time;
```
chapulina, line 468: We already spoke about this, but I'll just mention here for completeness that we could use the time from the message's header if that's present.

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,538 @@
...
+          mainChartID = Id;
+      }
+  }
+
+  Connections {
+      target: PlottingIface
+      onPlot : function (_chart, _fieldID, _x, _y)
+      {
+          charts[_chart].appendPoint(_fieldID, _x, _y);
+      }
```
chapulina, line 238: I traced the problem that we're having in seeing the plotted points to the way these functions are specified. Our version of Qt/QML doesn't seem to support this format: `slotName: function(params){}`.

Instead, this worked for me and I was able to see plotted points:

```diff
diff --git a/include/ignition/gui/qml/PlottingInterface.qml b/include/ignition/gui/qml/PlottingInterface.qml
index 9b88bf7..de776e5 100644
--- a/include/ignition/gui/qml/PlottingInterface.qml
+++ b/include/ignition/gui/qml/PlottingInterface.qml
@@ -230,12 +230,14 @@ Rectangle
       }
   }
 
+   function handlePlot(_chart, _fieldID, _x, _y)
+   {
+     charts[_chart].appendPoint(_fieldID, _x, _y);
+   }
+
   Connections {
       target: PlottingIface
-      onPlot : function (_chart, _fieldID, _x, _y)
-      {
-          charts[_chart].appendPoint(_fieldID, _x, _y);
-      }
+      onPlot: handlePlot(_chart, _fieldID, _x, _y)
       onMoveChart :function() {
           Object.keys(charts).forEach(function(key) {
               charts[key].moveChart();
```

Can you change all connections to use this style?

include/ignition/gui/qml/PlottingInterface.qml
AmrElsersy: ok I used the ToolButton in other buttons that has Rectangle + Text + MouseArea
but this one specifically i found that i will write a bigger code to style the ToolButton as i want than the Rectangle + Text + MouseArea :D, so i keep it

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,531 @@
...
+      anchors.top: addBtn.bottom
+      anchors.margins: 10
+      onHoveredChanged: (opacity === 1) ? opacity = 0.8 : opacity = 1;
+
+      icon.width: parent.width
+      icon.height: parent.height
+      icon.color: (Material.theme == Material.Light) ? Material.color(Material.Grey, Material.Shade700)
+                                                     : Material.color(Material.Grey, Material.Shade400)
+
+      icon.source: "images/export_icon.png"
```
claireyywang, line 264: I'm experiencing issue loading `icon` with Qt 5.9
```
file:///home/claire/gsoc_dome_ws/src/ign-gui/include/ignition/gui/qml/PlottingInterface.qml:259 Cannot assign to non-existent property "icon"
```
With a bit of research, I found some people say `icon` doesn't work prior to Qt5.10 https://github.com/moonlight-stream/moonlight-qt/issues/53#issuecomment-417181817. I was wondering if this is another modern feature that needs to be adjusted

claireyywang: here is a potential solution https://github.com/moonlight-stream/moonlight-qt/commit/0467c2f893cb35b7cab941d7a5d258f1e5731730#diff-81df80a5e56b87fa3890edbb06254763R75 

They used the icon image instead to make it compatible with Qt5.9

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,556 @@
...
+              options: FolderDialog.ShowDirsOnly
+
+              onAccepted: {
+                  if (exportBtn.currentText == "CSV")
+                  {
+                      var success = exportApp.exportCSV(folder);
+                      if (success)
+                          exportApp.close();
+                  }
+                  else if (exportBtn.currentText == "PDF")
```
chapulina, line 496: Let's remove all mentions of PDF because it's not supported yet. They should all be added at the same time once the feature is added.

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,556 @@
...
+  property var charts: ({})
+  property int mainChartID: 1
+
+  property int idIncrementor: 0
+
+  property bool multiChartsMode: false
+
+  // Horizonal Layout to hold multi charts (small charts)
+  Rectangle {
+      id: rowCharts
```
chapulina, line 44: Could you change the indentation of this and all QML files to be 2 spaces consistently?

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,556 @@
...
+  Layout.minimumHeight: 600
+  anchors.fill: parent
+  color: (Material.theme == Material.Light) ? Material.color(Material.Grey,Material.Shade100) : "transparent"
+
+  property var charts: ({})
+  property int mainChartID: 1
+
+  property int idIncrementor: 0
+
+  property bool multiChartsMode: false
```
chapulina, line 40: Please document all properties

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,556 @@
...
+import "qrc:/qml"
+
+Rectangle
+{
+  id : main
+
+  Layout.minimumWidth: 600
+  Layout.minimumHeight: 600
+  anchors.fill: parent
+  color: (Material.theme == Material.Light) ? Material.color(Material.Grey,Material.Shade100) : "transparent"
```
chapulina, line 33: The transparent background on dark mode makes the UI elements look disconnected. How about just using `Material.background` here, which should change appropriately according to the theme?

include/ignition/gui/qml/PlottingInterface.qml
```diff
@@ -0,0 +1,556 @@
...
+          }
+      }
+
+      ToolTip.text: "Add Chart"
+      ToolTip.delay: 500
+      ToolTip.timeout: 1000
+      ToolTip.visible: mouseAddBtn.containsMouse
+  }
+
+  function addChart()
```
chapulina, line 132: Also remember to document all functions. And I recommend moving all functions and properties to the top, before any UI items.

src/plugins/plotting/TransportPlotting.hh
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_GUI_PLUGINS_TRANSPORTPLOTTING_HH_
+#define IGNITION_GUI_PLUGINS_TRANSPORTPLOTTING_HH_
+
+#include <ignition/gui/Plugin.hh>
+#include <ignition/gui/PlottingInterface.hh>
+#include <memory>
```
claireyywang, line 22:
```suggestion
#include <ignition/utilities/SuppressWarning.hh>

#include <memory>
```

src/plugins/plotting/TransportPlotting.hh
```diff
@@ -0,0 +1,53 @@
...
+  public: TransportPlotting();
+
+  /// \brief Destructor
+  public: ~TransportPlotting();
+
+  // Documentation inherited
+  public: void LoadConfig(const tinyxml2::XMLElement *) override;
+
+  /// \brief Interface with the UI to Handle Transport Plotting
+  private: std::unique_ptr<PlottingInterface> dataPtr;
```
claireyywang, line 47:
```suggestion
  IGN_UTILS_WARN_IGNORE__DLL_INTERFACE_MISSING
  private: std::unique_ptr<PlottingInterface> dataPtr;
  IGN_UTILS_WARN_RESUME__DLL_INTERFACE_MISSING
```

PR #91, opened 2020-07-07
[Bazel] Fix various unit tests

mjcarroll:


src/plugins/BUILD.bazel
```diff
@@ -66,24 +64,11 @@ qt_cc_binary(
...
-        "-Iexternal/qt/QtGui",
-        "-Iexternal/qt/QtQuick",
-        "-Iexternal/qt/QtQuickControls2",
-        "-Iexternal/qt/QtWidgets",
-        "-Iexternal/qt/QtQml",
     ],
     deps = [
         "//ign_plugin/register",
         "//ign_transport",
-        "//ign_rendering",
```
JShep1, line 86: The build actually fails for me if this dependency isn't included

PR #98, opened 2020-08-01
 Fix tests on actions 

chapulina:
Also update the action style.


chapulina:
+45.62% test coverage at once :smile:

.github/ci/after_make.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh -l
+
+set -x
+
+# Install (needed for some tests)
+make install
```
mjcarroll, line 6: Doesn't have to be addressed here, but I think this is resolved in the bazel branch, I'll port it over.

src/plugins/CMakeLists.txt
```diff
@@ -81,10 +81,7 @@ function(ign_gui_add_plugin plugin_name)
     PRIVATE_LINK_LIBS ${ign_gui_add_plugin_PRIVATE_LINK_LIBS} ignition-plugin${IGN_PLUGIN_VER}::register
   )
 
-message(STATUS "ddddddddddddddddddddddddddddddddddddddddd")
```
mjcarroll, line 84: Beautiful.

.github/ci/after_make.sh
chapulina: Oh sweet, several libraries need this fix

chapulina:
> were you going to add focal as a platform or is it more work than expected?

We haven't released Blueprint debs for Focal, so we'd need to build dependencies from source. So my plan is to add Focal from Citadel :crossed_fingers:

.github/ci/after_make.sh
mjcarroll: Hoping to do away with it across the stack :crossed_fingers:

PR #101, opened 2020-08-04
3 ➡️ 4

chapulina:


Sarath18:
In case you missed it, we should incorporate changes from #102 as we are introducing KeyPublisher and Minimize GUI plugins to ign-gui4 in this PR.

chapulina:
> we should incorporate changes from #102

Done in f21238d 

> codecheck

Depending on the version of codecheck we may get different warnings, so at the moment, we're trying to satisfy the version on CI (it runs on actions and Jenkins for Ubuntu), which is currently happy. But in any case, I fixed the missing include in 
f36ce44 . The unmatched suppression are trickier though :confused:

JShep1:
> Depending on the version of codecheck

Perhaps it's time I update :thinking:

chapulina:
> Perhaps it's time I update

I suspect you may have a newer version :monocle_face: The one on CI didn't catch the missing include

PR #76, opened 2020-06-10
Disable failing tests

chapulina:
Having constant unstable CI makes it harder to notice when new test failures are introduced. The goal is to have green CI across the board and keep it that way.

This PR is just disabling known test failures, which have been listed on #75 . We should go back and fix these asap instead of just leaving them there forever.

Issue #75, opened 2020-06-10
Re-enable skipped tests for OSX and Windows


chapulina:
I'll be disabling several tests that are known to fail on OSX and Windows so we can go back to green CI. We should go back to fix these asap.

## OSX / MacOS / Apple

- [ ] `UNIT_Publisher_TEST`
    * It segfaults even before printing any Ignition console messages. It's very similar to `UNIT_WorldControl_TEST`, but that one passes on Dome.
- [ ] `INTEGRATION_Examples_TEST` - `ExampleTest.Configs`
- [ ] `INTEGRATION_ExamplesBuild_TEST`- `Plugins/ExamplesBuild.Build/0`
- [ ] `UNIT_Publisher_TEST` - `PublisherTest.Load`
- [ ] `UNIT_TopicViewer_TEST` - `TopicViewerTest.Load`
- [ ] `UNIT_WorldControl_TEST` - `WorldControlTest.Load`
- [ ] `UNIT_WorldStats_TEST` - `WorldStatsTest.Load`
- [ ] `UNIT_MainWindow_TEST`- `MainWindowTest.OnAddPlugin`
- [ ] `UNIT_ign_TEST` - `CmdLine.list`
    <details>

    ```
    9: /Users/jenkins/workspace/ignition_gui-ci-ign-gui2-homebrew-amd64/ign-gui/src/ign_TEST.cc:57: Failure
    9: Expected: (output.find("TopicEcho")) != (std::string::npos), actual: 18446744073709551615 vs 18446744073709551615
    9: /Users/jenkins/workspace/ignition_gui-ci-ign-gui2-homebrew-amd64/ign-gui/src/ign_TEST.cc:58: Failure
    9: Expected: (output.find("Publisher")) != (std::string::npos), actual: 18446744073709551615 vs 18446744073709551615
    ```
    </details>

## Windows

* Timeout with `unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.`
    - [ ] `UNIT_Application_TEST` - `ApplicationTest.Constructor`
    - [ ] `UNIT_MainWindow_TEST`- `MainWindowTest.Constructor`
    - [ ] `UNIT_Publisher_TEST` - `PublisherTest.Load`
    - [ ] `UNIT_Plugin_TEST` - `PluginTest.DeleteLater`
    - [ ] `UNIT_TopicViewer_TEST` - `TopicViewerTest.Load`
[...]

PR #103, opened 2020-08-05
Disable TopicViewer_TEST on Windows and macOS

chapulina:
See #75.

I thought we could wait for #76 to be merged forward to disable the test, but on Windows the compilation was broken.

### DIFF
--- a/src/plugins/topic_viewer/TopicViewer_TEST.cc
+++ b/src/plugins/topic_viewer/TopicViewer_TEST.cc
@@ -37,6 +37,8 @@ using namespace ignition;
 using namespace gui;
 using namespace plugins;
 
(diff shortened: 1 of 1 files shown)


claireyywang:
There are some warnings on ubuntu CI related to the original PR https://build.osrfoundation.org/job/ignition_gui-ci-pr_any-ubuntu_auto-amd64/577/gcc/new/ERROR/, not sure if that'll break the buildfarm though

chapulina:
> There are some warnings on ubuntu CI related to the original PR

Sorry, I had missed this comment. I think those warnings are only showing up now because of the new Jenkins plugin. I'm addressing them in https://github.com/ignitionrobotics/ign-gui/pull/101

PR #111, opened 2020-08-19
Fix running plugins with bad-er world name

chapulina:
This is a redo of #107, which was reverted in #110.

It now also covers the use case where the `worldNames` property is not set.


nkoenig:
I think there might still be a problem. I see this error:

```
[GUI] [Wrn] [WorldControl.cc:173] Ignoring topic [/world/simple_cave_01/stats], world name different from []. Fix or remove your <stats_topic> tag.
Topic [/world//stats] is not valid.
[GUI] [Err] [WorldControl.cc:191] Failed to subscribe to [/world//stats]
```

chapulina:
> I see this error:

Thanks, you found the one case that I didn't test :see_no_evil: Fixed: 4bfc316c01a40abb194fed3964f8a441f314bdf8

PR #116, opened 2020-08-26
Using std::chrono instead of common::Time

ahcorde:
This PR is related with this issue https://github.com/ignitionrobotics/ign-common/issues/61. The idea it's to deprecate the class `common::Time` in favor of `std::chrono`.

Related with 
 - https://github.com/ignitionrobotics/ign-math/pull/150

Should I deprecate the methods/functions that use common:Time ?


chapulina:
> Should I deprecate the methods/functions that use common:Time ?

We should check who's using those functions and make sure to update them before committing to deprecating them. I think here the only case is for the `convert` functions.

include/ignition/gui/Conversions.hh
```diff
@@ -91,12 +91,14 @@ namespace ignition
...
     IGNITION_GUI_VISIBLE
+    IGN_DEPRECATED(4)
     common::Time convert(const msgs::Time &_t);
 
     /// \brief Convert an ignition::common::Time to an ignition::msgs::Time
     /// \param[in] _t The time to convert
     /// \return An ignition::msgs::Time object
     IGNITION_GUI_VISIBLE
+    IGN_DEPRECATED(4)
     msgs::Time convert(const common::Time &_t);
```
chapulina, line 102: How about adding new functions that use `chrono` instead of `ignition::common`? Whoever is using these functions now will need an alternative.

chapulina: Oh now it's possible to use the ones from `ign-msgs` added in https://github.com/ignitionrobotics/ign-msgs/pull/81, I'll just mention that in the migration guide.

.github/ci/packages.apt
```diff
@@ -25,3 +25,4 @@ qtbase5-dev
 qtdeclarative5-dev
 qtquickcontrols2-5-dev
 xvfb
+python3-empy
```
chapulina, line 28: Why this is needed?

chapulina: Gave a try at removing it on d9bb607

chapulina (approved):

PR #124, opened 2020-09-13
Fix Plotting Crash

AmrElsersy:
Fix Plotting Crash due to closing Plotting ign-gui plugin

@chapulina 
@claireyywang


src/PlottingInterface.cc
```diff
@@ -80,19 +80,19 @@ class TransportPrivate
 class PlottingIfacePrivate
 {
   /// \brief Responsible for transport messages and topics
-  public: Transport *transport;
+  public: Transport transport;
 
   /// \brief current plotting time
-  public: double time;
+  public: double *time {new double};
```
chapulina, line 86: Is this variable needed? Can we just have `plottingTimeRef`?

src/plugins/plotting/TransportPlotting.cc
```diff
@@ -21,7 +21,8 @@ using namespace ignition;
 using namespace gui;
 using namespace plugins;
 
-TransportPlotting::~TransportPlotting() {}
+TransportPlotting::~TransportPlotting() {
```
chapulina, line 24:
```suggestion
TransportPlotting::~TransportPlotting()
{
```

chapulina (changes requested):

src/PlottingInterface.cc
AmrElsersy: yeah great idea :D 
i modified it

chapulina (approved):
LGTM! I just pushed a little change in d65abeed042004e8d417d70541b5afb5a11bfd5d to use `make_shared` instead of `new`, which is slightly more efficient.

PR #125, opened 2020-09-17
Small tweaks to the plot plugin

chapulina:
Just a couple of tweaks that I noticed while reviewing https://github.com/ignitionrobotics/ign-gazebo/pull/270/:

* The check in `Chart.qml` is needed to avoid warnings after a field is deleted from a plot
* Changed the title of the transport plotting plugin so it's different from the one in `ign-gazebo`.

@AmrElsersy , mind taking a look?


claireyywang (approved):
LGTM

PR #126, opened 2020-09-23
Fix qt5 anchor warnings

luca-della-vedova:
Same as the [ign-gazebo PR](https://github.com/ignitionrobotics/ign-gazebo/pull/363), fixes the QT5 warnings at simulation starting time caused by the WorldControl plugin:

```
[ign-8] [GUI] [Wrn] [Application.cc:649] [QT] file::/WorldControl/WorldControl.qml:88:3: QML RoundButton: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
```


chapulina (approved):
Thanks!

PR #133, opened 2020-09-30
⬆️  4.0.0: Dome stable release

chapulina:
Comparison to `ign-gui3`: https://github.com/ignitionrobotics/ign-gui/compare/ign-gui3...ign-gui4

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,44 @@
 
 ### Ignition Gui 4.X.X
 
-### Ignition Gui 4.0.0 (20XX-XX-XX)
+### Ignition Gui 4.X.X (20XX-XX-XX)
+
+### Ignition Gui 4.0.0 (2020-09-29)
+
+1. GitHub migration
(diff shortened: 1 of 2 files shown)


Changelog.md
```diff
@@ -2,7 +2,44 @@
 
 ### Ignition Gui 4.X.X
 
-### Ignition Gui 4.0.0 (20XX-XX-XX)
+### Ignition Gui 4.X.X (20XX-XX-XX)
```
adlarkin, line 5: Perhaps this is a misunderstanding from my end, but why are we moving from `4.0.0` to `4.X.X`? `4.0.0` is still being used a few lines below.

adlarkin (approved):
Looks good to me! Just one minor question.

Changelog.md
chapulina: Oh, this is just a placeholder here for the next minor / patch release. People will often add entries below the `N.X.X` before the release

PR #134, opened 2020-09-30
Add scrollable indicator for plugin menu

Sarath18:
Reference to issue [ign-gazebo #368](https://github.com/ignitionrobotics/ign-gazebo/issues/368)


chapulina (approved):
Yup, it's always there now, thanks!

scroll_menu

PR #135, opened 2020-10-01
Floating and standalone plugins respect minimum dimensions

Sarath18:
* Set parentItem after resizing the dialog

* Create minimum height and width card properties

* Set height and width of card on adding a child plugin

* Enforce minimum dimensions on floating plugin resize


Sarath18:
Instead of setting height and width on Component.onCompleted I set the height and width of the card pane when the child plugin is added. This way we don't require to change any of the plugins.

chapulina (approved):
Great improvements, thanks for the fix!

PR #138, opened 2020-10-08
⬆️  Bump main to 5.0.0~pre1

chapulina:
The `ign-gui4` branch was created off `main` for the Dome release.

Now bumping `main` to version 5 so it may receive breaking changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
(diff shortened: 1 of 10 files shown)


j-rivero (approved):
I modified the badge for the Windows CI job. There are references in tutorials for 4.0 version of API docs. I think that we are not generating documentation for main/master branch, is this correct?

chapulina:
>  There are references in tutorials for 4.0 version of API docs. I think that we are not generating documentation for main/master branch, is this correct?

Nice catch, yes that's correct. I don't think we have a better link to point these tutorials to now. This is not ideal because once we release Edifice we may forget to update these links :confused:

PR #144, opened 2020-11-04
Resolve updated codecheck issues

adlarkin:
This PR is meant to resolve new codecheck issues introduced by https://github.com/ignitionrobotics/ign-cmake/pull/117.

From my understanding, the plan is to forward port the fixes here to all other ignition versions once https://github.com/ignitionrobotics/ign-cmake/pull/117 is merged.

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

### DIFF
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -15,8 +15,14 @@
  *
 */
 
-#include <QQuickImageProvider>
+#include "ImageDisplay.hh"
+
+#include <algorithm>
 #include <iostream>
+#include <string>
+#include <vector>
+
+#include <QQuickImageProvider>
 
 #include <ignition/common/Console.hh>
 #include <ignition/common/Image.hh>
@@ -24,7 +30,6 @@
 #include <ignition/transport/Node.hh>
 
 #include "ignition/gui/Application.hh"
-#include "ImageDisplay.hh"
 
 namespace ignition
 {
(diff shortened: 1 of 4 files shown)


src/plugins/image_display/ImageDisplay.cc
```diff
@@ -15,16 +15,21 @@
...
+
+#include <QQuickImageProvider>
 
 #include <ignition/common/Console.hh>
 #include <ignition/common/Image.hh>
 #include <ignition/plugin/Register.hh>
 #include <ignition/transport/Node.hh>
 
 #include "ignition/gui/Application.hh"
-#include "ImageDisplay.hh"
```
chapulina, line 27: This change makes the local code checker fail:

```
./src/plugins/image_display/ImageDisplay.cc:25:  Found C system header after C++ system header. Should be: ImageDisplay.h, c system, c++ system, other.  [build/include_order] [4]
```

chapulina (changes requested):

src/plugins/image_display/ImageDisplay.cc
adlarkin: I believe the error is actually coming from `#include <QQuickImageProvider>` (it looks like the local checker interprets this as a C system header). I'm working on a fix.

adlarkin: Should be fixed now in https://github.com/ignitionrobotics/ign-gui/pull/144/commits/3b1494cf12ae07bc1f53429fca8cc1ccd48e4ce1.

chapulina (approved):
Looks good now :+1:

PR #148, opened 2020-11-20
Add gazebo gui events

JShep1:

### DIFF
--- a/include/ignition/gui/GuiEvents.hh
+++ b/include/ignition/gui/GuiEvents.hh
@@ -45,6 +45,8 @@ namespace ignition
       };
 
       /// \brief The class for sending and receiving custom snap value events.
+      /// This event is used in the Transform Control plugin tool when the
+      /// user manually alters their snapping values.
       class SnapIntervals : public QEvent
       {
         /// \brief Constructor
@@ -81,7 +83,7 @@ namespace ignition
         }
   
         /// \brief The QEvent representing a snap event occurrence.
-        static const QEvent::Type kType = QEvent::Type(QEvent::User);
+        static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser - 1);
   
         /// \brief XYZ snapping values in meters, these values must be positive.
         private: math::Vector3d xyz;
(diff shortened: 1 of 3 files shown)


include/ignition/gui/GuiEvents.hh
```diff
@@ -43,6 +43,158 @@ namespace ignition
...
+        {
+          return this->modelSdfString;
+        }
+  
+        /// \brief The sdf string of the model to be previewed.
+        std::string modelSdfString;
+      };
+  
+      /// \brief Event called to spawn a preview resource, which takes the path
+      /// to the SDF file. Used by plugins that spawn resources.
```
chapulina, line 123: Ignition GUI doesn't know about SDF, so we could just say "path to a file".

include/ignition/gui/GuiEvents.hh
```diff
@@ -43,6 +43,158 @@ namespace ignition
...
+        /// these values must be positive.
+        private: math::Vector3d scale;
+      };
+
+      /// \brief Event called to spawn a preview model.
+      /// Used by plugins that spawn models.
+      class SpawnPreviewModel : public QEvent
+      {
+        /// \brief Constructor
+        /// \param[in] _modelSdfString The model's SDF file as a string.
```
chapulina, line 102: Same here about Ignition GUI not knowing about SDF, let's say something like "the model's description as a string"

include/ignition/gui/GuiEvents.hh
```diff
@@ -43,6 +43,158 @@ namespace ignition
         /// \brief Unique type for this event.
         static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser);
       };
+
+      /// \brief The class for sending and receiving custom snap value events.
```
chapulina, line 47: I wonder if we can improve this documentation to provide more context. For a user who doesn't know about the snap functionality, the current documentation can be confusing.

include/ignition/gui/GuiEvents.hh
```diff
@@ -43,6 +43,158 @@ namespace ignition
...
+  
+        /// \brief Get the scale snapping values.
+        /// \return The scale snapping values.
+        public: math::Vector3d Scale() const
+        {
+          return this->scale;
+        }
+  
+        /// \brief The QEvent representing a snap event occurrence.
+        static const QEvent::Type kType = QEvent::Type(QEvent::User);
```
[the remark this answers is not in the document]
JShep1: be5a482d8da3da972dff28a187723c805378d367

include/ignition/gui/GuiEvents.hh
JShep1: be5a482d8da3da972dff28a187723c805378d367

include/ignition/gui/GuiEvents.hh
JShep1: be5a482d8da3da972dff28a187723c805378d367

include/ignition/gui/GuiEvents.hh
JShep1: be5a482d8da3da972dff28a187723c805378d367

include/ignition/gui/GuiEvents.hh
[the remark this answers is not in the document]
JShep1: I updated the events to have different ID's.  I was thinking we could also have them share the same ID's and then try to cast as both the `ignition::gui` and `ignition::gazebo::gui` events if that type is found, since they technically represent the same event as well.  Not sure which approach you think would be better.  On one hand we have two separate events for the same functionality, on the other hand, we have to add a little more redundancy into the event casting.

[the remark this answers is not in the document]
chapulina: Oh interesting idea with the casting. One thing I think could be risky is trying to keep both in sync though... In the past we've tried to keep enums in sync across libraries and it can become a headache :confused: I recognize that the situation here would just be a temporary transition though.

chapulina (changes requested):
The code checker is not happy:

https://github.com/ignitionrobotics/ign-gui/pull/148/checks?check_run_id=1433264752

---

This looks good to me. Just to add more context for future readers, the idea here is to define events in `ign-gui` so they can be reused by multiple downstream projects like `ign-gazebo` and `ign-rviz`. This is also related to https://github.com/ignitionrobotics/ign-gui/issues/68.

We should deprecate [ignition::gazebo::gui::GuiEvents](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/include/ignition/gazebo/gui/GuiEvents.hh) in favor of these events.

I also just realized we don't have unit tests for the events. This is especially important in this case, because there aren't plugins using these events that can be tested. @JShep1, sorry to increase the scope of this PR, but do you think you could add some tests? Let me know how I can help.

chapulina:
I opened #150 with some suggestions.

JShep1:
> I opened #150 with some suggestions.

Thank you!  I just merged your changes in

chapulina (approved):

PR #150, opened 2020-12-10
Suggestions to #148

chapulina:
* Address codechecker warnings
* Added tests
* Pass and return `const &`
* Changed the snap API to `Position` / `Rotation` / `Scale`, to better match APIs of other Ignition libraries, like math, rendering, common...
* I think the "preview" concept is very specific to Gazebo and its server-client architecture. Other applications may be interested in using this event to just spawn anything into a scene. I think we'll only really know once we start making use of that event in `Scene3D`.


JShep1 (approved):
LGTM, thanks for the tweaks

PR #153, opened 2020-12-10
Change deprecated Qt::MidButton

chapulina:
I noticed some new warnings on macOS:

https://build.osrfoundation.org/job/ignition_gui-ci-pr_any-homebrew-amd64/709/clang/new/

`'MidButton' is deprecated: MidButton is deprecated. Use MiddleButton instead`

This is probably only showing on macOS because it uses a newer version of Qt.

In any case `Qt::MiddleButton` has been an alias for `Qt::MidButton` for a long time, [since at least Qt4](https://doc.qt.io/archives/qt-4.8/qt.html), so this should be a safe change on all platforms.


JShep1 (approved):
:100:

PR #154, opened 2020-12-10
Add right mouse events and tests

JShep1:
Used by https://github.com/ignitionrobotics/ign-gazebo/pull/458/

### DIFF
--- a/include/ignition/gui/GuiEvents.hh
+++ b/include/ignition/gui/GuiEvents.hh
@@ -228,15 +228,15 @@ namespace ignition
         private: math::Vector3d point;
       };
 
-      /// \brief Event which is called to enable or disable the right click
-      /// dropdown menu. This is primarily used by plugins which also use
-      /// the right click to cancel any actions currently in progress.
-      class RightClickDropdownMenu : public QEvent
+      /// \brief Event which is called to enable or disable the dropdown menu.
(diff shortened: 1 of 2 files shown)


include/ignition/gui/GuiEvents.hh
```diff
@@ -201,6 +201,60 @@ namespace ignition
...
+          return this->point;
+        }
+
+        /// \brief The 3D point that the user clicked within the scene.
+        private: math::Vector3d point;
+      };
+
+      /// \brief Event which is called to enable or disable the right click
+      /// dropdown menu. This is primarily used by plugins which also use
+      /// the right click to cancel any actions currently in progress.
```
chapulina, line 233: So this event needs to be used together with `RightClickToScene`, right? I'm wondering if it wouldn't be enough to call it `DropdownMenu`, since it doesn't carry any information about the click.

chapulina: I just looked at how it's being used and now I think a better name may be `DropdownMenuEnabled`

JShep1: True! Updated here 1b8204e841affe2b0ca602e4e215ad557df2a020

chapulina (approved):

PR #157, opened 2020-12-18
Fix codecheck

chapulina:
Satisfy new codechecker when doing `make codecheck`. That's tested on GitHub actions CI.


adlarkin (approved):

chapulina:
Codecheck passed on actions and jenkins! Merging!

PR #161, opened 2020-12-24
Add Windows installation

mabelzhang:
Partially addresses https://github.com/ignitionrobotics/docs/issues/117

Moved Documentation and Testing sections from the installation tutorial back to README - looks like a mistake that they were moved before. These two sections stayed in the README for other libraries.

One of the dependencies, `libignition-rendering`, is not yet on conda-forge and needs to be installed from source.
`libignition-gui` is also not on conda-forge, so we don't have binary install instructions for Windows in this PR.
Both libraries need to be added to conda-forge.


tutorials/01_install.md
```diff
@@ -87,58 +90,66 @@ Build and install as follows:
...
     ```
 
-# Testing
-
-Follow these steps to run tests and static code analysis in your clone of this repository.
-
-1. Follow the [source install instruction](#source-install).
-
-2. Run tests.
+    Before ``ign-rendering`` becomes available on conda-forge, we need to build it from source and specify the path containing ``ignition-rendering<#>-config.cmake`` and ``ignition-rendering<#>-ogre-config.cmake`` in ``CMAKE_PREFIX_PATH``. That path could be ``ign-rendering-install-path\lib\cmake``, for example.
```
JShep1, line 144:
```suggestion
    As``ign-rendering`` is not available on conda-forge, we need to build it from source and specify the path containing ``ignition-rendering<#>-config.cmake`` and ``ignition-rendering<#>-ogre-config.cmake`` in ``CMAKE_PREFIX_PATH``. That path could be ``ign-rendering-install-path\lib\cmake``, for example.
```

JShep1 (commented):
Small verbiage nit

chapulina:
> Moved Documentation and Testing sections from the installation tutorial back to README - looks like a mistake that they were moved before

I think I did that on purpose on https://github.com/ignitionrobotics/ign-gui/pull/127 because they seemed out of place in the README. A README user had no context for generating docs or running tests... But I think that you are adding enough context in this PR. Maybe we can consider moving those to their own tutorials in the future. I think the README could be more focused on highlighting features and quick onboarding, and specific development details can be left to tutorials.

---

Thanks for the PR. Just curious, did you try running `ign gui`? Does it work? How about the 3D scene plugin (`ign gui -c examples/config/scene3d.config`)?

JShep1:
Closing to retarget at Citadel. https://github.com/ignitionrobotics/ign-gui/pull/168

JShep1:
> Thanks for the PR. Just curious, did you try running `ign gui`? Does it work? How about the 3D scene plugin (`ign gui -c examples/config/scene3d.config`)?

For visibility, I tested this out by attempting to run a few examples and Qt crashed every time, for what it's worth, here is the error message in case this may be relevant later:
```
(test_ws) C:\Users\asus\test-ws\src\ign-gui\examples\standalone\window\build\Release>window.exe
Hello, GUI!
qt.qpa.plugin: Could not find the Qt platform plugin "windows" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

Not sure if this may be a local issue or if reinstalling Qt would actually work, it could also be some sort of issue with the version of Qt that `conda-forge` brings in.

PR #163, opened 2020-12-30
Generate valid topics depending on world name

chapulina:
Part of https://github.com/ignitionrobotics/ign-gazebo/issues/239 is making sure that all topics can handle entity names with spaces. This can be the case for the world.

This makes use of the `AsValidTopic` function to convert world names to valid ones whenever possible.


chapulina:
There's little value in getting this into Citadel because the string components serialization in Ignition Gazebo isn't working properly with whitespaces. This is fixed on Dome though, so I retargeted the PR on #164. We can backport that to Citadel in the future if we see the need.

PR #164, opened 2020-12-30
Generate valid topics depending on world name

chapulina:
Redo of #163 targeting Dome.

Part of https://github.com/ignitionrobotics/ign-gazebo/issues/239 is making sure that all topics can handle entity names with spaces. This can be the case for the world.

This makes use of the `AsValidTopic` function to convert world names to valid ones whenever possible.


JShep1 (approved):
LGTM with all tests passing

PR #170, opened 2021-01-28
Screenshot plugin

jennuine:
Closes https://github.com/ignitionrobotics/ign-gui/issues/95

This plugin takes screenshots similar to Gazebo-classic by clicking the camera icon. The plugin uses a screenshot service that saves an image of the current 3D scene from the user camera. The default directory of saved screenshots is `/home/.ignition/gazebo/pictures` but users can change the directory by selecting the folder icon.

~~This can be used along with ign-gazebo's screenshot service in https://github.com/ignitionrobotics/ign-gazebo/pull/588~~

EDIT: depends on https://github.com/ignitionrobotics/ign-gazebo/pull/598

Peek 2021-01-27 17-55


src/plugins/scene3d/Scene3D.hh
```diff
@@ -72,6 +75,13 @@ namespace plugins
     public: virtual void LoadConfig(const tinyxml2::XMLElement *_pluginElem)
         override;
 
+    /// \brief Callback for saving a screenshot (from the user camera) request
+    /// \param[in] _msg Request message of the saved file path
+    /// \param[in] _res Response data
+    /// \return True if the request is received
+    private: bool OnScreenshot(const msgs::StringMsg &_msg,
+        msgs::Boolean &_res);
```
chapulina, line 83: See suggestions on #171 on how to keep all the logic within the Screenshot plugin.

src/plugins/screenshot/Screenshot.cc
```diff
@@ -0,0 +1,130 @@
...
+
+/////////////////////////////////////////////////
+void Screenshot::OnChangeDirectory(const QString &_dirUrl)
+{
+  QString newDir = QUrl(_dirUrl).toLocalFile();
+  this->dataPtr->directory = newDir.toStdString();
+
+  // Update tooltip text
+  App()->Engine()->rootContext()->setContextProperty("ScreenshotDirectory",
+      newDir);
```
chapulina, line 125: We should avoid using these "global" properties, because there could be collisions with other plugins. I suggest adding a property to this plugin.

See the `msgType` property of the `Publisher` plugin as an example:

https://github.com/ignitionrobotics/ign-gui/blob/9140459ed06a65e3045354064edfb7e4cc8b9e72/src/plugins/publisher/Publisher.hh#L51-L57

https://github.com/ignitionrobotics/ign-gui/blob/9140459ed06a65e3045354064edfb7e4cc8b9e72/src/plugins/publisher/Publisher.hh#L96-L107

chapulina: Another suggestion, you could even use the same approach to store the final timestamped path as a property, and display that on the dialog. That should also make the dialog only pop up after the image has actually been saved.

chapulina (commented):
Works for me! Some suggestions below:

jennuine:
Currently looking into an issue where the colors in the the saved screenshot do not match what is shown in ignition. Here are some examples, the image on the left is ignition and on the right is the saved screenshot:

Screenshot from 2021-02-01 15-24-52

Screenshot from 2021-02-01 15-27-13

chapulina:
Here's a fix for the swapped colors: https://github.com/ignitionrobotics/ign-common/pull/162

jennuine:
> Also, do you mind adding some basic tests? Maybe just instantiating the plugin, requesting the service, and checking that a file was created?

Added a simple test that only instantiates plugin e8c3cd231e9495c85bfa14da8aa78e6226af6799. The code that loads the Scene3D plugin and requests the service is commented out because currently unable to load Scene3D from another plugin.

src/plugins/screenshot/Screenshot.cc
jennuine: Done 9b5e3db77216ffccca5a23b53fe101493e9f2f08

src/plugins/scene3d/Scene3D.hh
```diff
@@ -18,6 +18,9 @@
 #ifndef IGNITION_GUI_PLUGINS_SCENE3D_HH_
 #define IGNITION_GUI_PLUGINS_SCENE3D_HH_
 
+#include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/stringmsg.pb.h>
```
chapulina, line 22: Can these be removed?

src/plugins/screenshot/Screenshot.cc
```diff
@@ -0,0 +1,276 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include "Screenshot.hh"
+
+#include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/stringmsg.pb.h>
```
chapulina, line 20: These `ign-msgs` includes introduce Windows warnings. We can suppress the warnings like this:

https://github.com/ignitionrobotics/ign-gui/blob/10affb2b101c333f0e4c194e79444ba9179f6b61/src/plugins/image_display/ImageDisplay.hh#L21-L28

The same goes for the test.

tutorials/08_screenshot.md
```diff
@@ -0,0 +1,32 @@
...
+It is possible to change the default directory to a different location, which will
+be shown in the example below.
+
+**Note**: This example is using Ignition Gazebo but is not required.
+The same steps can be done using `ign gui` with a 3D scene.
+
+First, go to the plugins menu on the top right (vertical ellipsis) then select
+'Screenshot':
+
+![Screenshot from dropdown menu](images/screenshot_menu.png)
```
chapulina, line 16: This way of adding images works on GitHub, but not after our documentation is uploaded to https://ignitionrobotics.org. We should use the `@image` syntax like [`ign-rendering` does for example](https://github.com/ignitionrobotics/ign-rendering/blame/ign-rendering4/tutorials/13_custom_scene_viewer.md#L43). 

We'll also need to add a line like this to `CMakeLists`:

`file(COPY ${CMAKE_SOURCE_DIR}/tutorials/images/ DESTINATION ${CMAKE_BINARY_DIR}/doxygen/html/images/)`

You can test the result locally following the documentation generation instructions [here](https://ignitionrobotics.org/api/gui/4.2/install.html).

src/plugins/scene3d/Scene3D.hh
jennuine: Yup, done a393787385e8b469bcaae8a5d44f8481717ed24c

src/plugins/screenshot/Screenshot.cc
jennuine: Thanks! a393787385e8b469bcaae8a5d44f8481717ed24c

tutorials/08_screenshot.md
jennuine: Thank you! a393787385e8b469bcaae8a5d44f8481717ed24c

src/plugins/screenshot/Screenshot.cc
```diff
@@ -0,0 +1,273 @@
...
+#include <string>
+
+#include <ignition/common/Console.hh>
+#include <ignition/common/Filesystem.hh>
+#include <ignition/common/Image.hh>
+#include <ignition/plugin/Register.hh>
+#include <ignition/rendering/Camera.hh>
+#include <ignition/rendering/RenderEngine.hh>
+#include <ignition/rendering/RenderingIface.hh>
+#include <ignition/rendering/Scene.hh>
```
chapulina, line 28: The last 2 Windows warnings are coming from `ign-rendering`. I'm not sure why they don't show up on upstream CI. In any case, we've been suppressing those warnings too, like this:

https://github.com/ignitionrobotics/ign-gui/blob/10affb2b101c333f0e4c194e79444ba9179f6b61/src/plugins/scene3d/Scene3D.cc#L35-L51

jennuine: Thank you! 8dd1d124427d92306da5fec432db4cfbc07deefd

chapulina:
I pushed c33970b directly to this branch so we suppress the `4251` warnings on all plugin files. I have no idea where the `ign-rendering` warnings are leaking through, but this should handle it. This warning is already suppressed on all `src` files that are not in plugins.

chapulina (approved):
Thanks for iterating, works great!

PR #171, opened 2021-01-29
Suggestions to #170

chapulina:
Refactoring suggestions so we keep all the screnshot logic within the `Screenshot` plugin. This way:

* There's no need to add any code to the scene, so no screenshot functionality is loaded unless the user wants to use it
* The same screenshot plugin is usable with `ign-gui` and `ign-gazebo` (needs https://github.com/ignitionrobotics/ign-gazebo/pull/598)

The tutorials introduced in https://github.com/ignitionrobotics/ign-gazebo/pull/596 provide more context on this refactoring.

### DIFF
--- a/src/plugins/scene3d/Scene3D.cc
+++ b/src/plugins/scene3d/Scene3D.cc
@@ -25,7 +25,6 @@
 #include <vector>
 
 #include <ignition/common/Console.hh>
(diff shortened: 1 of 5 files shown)


jennuine (approved):
Works and LGTM! Thank you @chapulina!

PR #176, opened 2021-02-05
Visualize single channel 8 bit image data

iche033:
Signed-off-by: Ian Chen <ichen@osrfoundation.org>

## Summary

Extended Image Display plugin to support visualizing single channel 8 bit unsigned integer data in a similar way to the 16 bit data. This is used by https://github.com/ignitionrobotics/ign-gazebo/pull/614 to visualize 8 bit thermal image data (see PR for screenshot)

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -345,6 +345,8 @@ void ImageDisplay::UpdateFromLInt16()
 /////////////////////////////////////////////////
 void ImageDisplay::UpdateFromLInt8()
 {
+  // todo(anyone) the code in this function is very similar to
+  // UpdateFromInt16 and UpdateFromFloat32. Consider merging these functions.
   unsigned int height = this->dataPtr->imageMsg.height();
   unsigned int width = this->dataPtr->imageMsg.width();
   QImage::Format qFormat = QImage::Format_RGB888;
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -160,6 +160,9 @@ void ImageDisplay::ProcessImage()
     case msgs::PixelFormatType::L_INT16:
       this->UpdateFromLInt16();
       break;
+    case msgs::PixelFormatType::L_INT8:
+      this->UpdateFromLInt8();
+      break;
     default:
     {
       ignwarn << "Unsupported image type: "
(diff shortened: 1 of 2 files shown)


src/plugins/image_display/ImageDisplay.hh
```diff
@@ -99,6 +99,9 @@ namespace plugins
     /// \brief Update from L_INT16
     private: void UpdateFromLInt16();
 
+    /// \brief Update from L_IN8T
+    private: void UpdateFromLInt8();
```
adlarkin, line 103: Does adding this method to a header file in this case not break API/ABI since this is a plugin?

src/plugins/image_display/ImageDisplay.cc
```diff
@@ -339,6 +342,60 @@ void ImageDisplay::UpdateFromLInt16()
   delete[] buffer;
 }
 
+/////////////////////////////////////////////////
+void ImageDisplay::UpdateFromLInt8()
```
adlarkin, line 346: A lot of the code in this method looks very similar to the other `UpdateFrom*` methods in this class. It would be nice to re-factor the `UpdateFrom*` methods later when we have more time to remove the code duplication. Maybe we can add a TODO note about this? You can assign me to it if you'd like.

src/plugins/image_display/ImageDisplay.hh
chapulina: These headers are not installed, so we don't need to worry about ABI.

But also note that adding a private function is ABI-safe. The problem is adding private member variables.

src/plugins/image_display/ImageDisplay.cc
iche033: added a todo note in c7b0fce and ticketed an issue https://github.com/ignitionrobotics/ign-gui/issues/177. Feel free to assign it to yourself once you get around to it, thanks.

adlarkin:
I think this is good to go - do we need to add some sort of test?

adlarkin (approved):
Since there are currently no tests for pixel formats with image display (https://github.com/ignitionrobotics/ign-gui/blob/a60cb7d61339b407ff3285ddbf16f618c4c4b3b9/src/plugins/image_display/ImageDisplay_TEST.cc), I think that this is fine as is. Perhaps we can add some tests later if people think that would be a worthwhile contribution.

PR #179, opened 2021-02-10
Fix version number

chapulina:
Update since #178, the release has already been triggered. This is just for bookkeeping.


adlarkin (approved):
Thanks, sorry for not getting this right the first time :grimacing:

PR #193, opened 2021-03-15
Merge 4 ➡️ main

scpeters:
# ➡️ Forward port

Port `ign-gui4` to `main`

Branch comparison: https://github.com/ignitionrobotics/ign-gui/compare/main...ign-gui4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)


Changelog.md
```diff
@@ -4,6 +4,35 @@
 
 ### Ignition Gui 4.X.X (20XX-XX-XX)
 
+### Ignition Gui 4.3.0 (2021-03-11)
```
adlarkin, line 7: I'm noticing some redundancies/duplication between the changelog for `4.3.0` and `3.5.0`. Would it be better to remove the duplicate changes in the `4.3.0` changelog and instead just say "also including all changes up to `3.5.0`"?

chapulina: We've been adding the "up to..." line to some versions for expediency, but the duplication makes it easier to find entries later, so I think we should default to that whenever time allows.

adlarkin: Sounds good. Everything here looks okay to me then :+1:

adlarkin (approved):

PR #194, opened 2021-03-16
Improve the height of plugins in the right split

chapulina:
# 🦟 Bug fix

## Summary

The height of plugins stacked on the right split doesn't always update in a reasonable way. This pull request improves the situation. The following situations use `examples/config/layout.config`.

* If the window is not large enough to fit all plugins, squish them all to their minimum height and make the split scrollable.
* If the window is large enough to fit all plugins, they are stretched to fill the entire height, and their proportion can be changed dragging the separator.
* Collapsed plugins behave consistently with the above
* Adding / removing plugins also behaves as expected

split1

split2

split3

## Checklist
- [ ] Added tests: No, because we don't have a good way of testing the view right now :confused: 
- [ ] ~~Updated migration guide (as needed)~~


scpeters:
I don't really get what's supposed to happen, and I am having trouble reproducing the behavior in your animations

scpeters:
> I don't really get what's supposed to happen, and I am having trouble reproducing the behavior in your animations

I see that the separator is not movable with the current version, but it is movable with this branch

scpeters:
> > I don't really get what's supposed to happen, and I am having trouble reproducing the behavior in your animations
> 
> I see that the separator is not movable with the current version, but it is movable with this branch

but I don't consistently see the auto-adjusting of size that you mention in the description

chapulina:
I should have mentioned the issue that this is trying to solve. Without this PR, the plugins are always squished to their minimum heights and don't expand to take up the free space. For example, see how the entity tree is small even though there's a lot of empty space under it:

image

This also means that the separators are never draggable without this PR, because the widgets are fixed to their minimum heights.

chapulina:
This issue was brought up here, see how the widget is squished there: https://github.com/ignitionrobotics/ign-gazebo/pull/534#pullrequestreview-564612159

scpeters (approved):

scpeters:
> * If the window is large enough to fit all plugins, they are stretched to fill the entire height, and their proportion can be changed dragging the separator.

I think I was misinterpreting what you meant by this. I see the same behavior as in your animations now, so it looks fine enough to me, even if I'm not sure what "stretched to fill the entire height" means

chapulina:
> I'm not sure what "stretched to fill the entire height" means

You can see in my gifs that unless the plugins are collapsed, there's never empty space below them. They are "stretched" to fill the entire column.

Thanks for the quick review!

PR #217, opened 2021-04-29
Consolidate scene3D with ign-gui

ahcorde:
# 🎉 New feature

This PR is part of the consolidation of Scene3D with ign-gazebo's GzScene3D https://github.com/ignitionrobotics/ign-gui/issues/137

Related to:
 - https://github.com/ignitionrobotics/ign-rendering/pull/311

## Summary

Make the `ign-gui` plugin extensible

I have create a new Plugin to test this new structure called Scene3DNew. I tried to launch it with `ign gui -s Scene3dNew` but for some reason it crash.

The other option is to use the windows example an change the plugin to Scene3DNew

Right now supports:

 - Models in the scene
 - Sky
 - moveTo
 - follow
 - Fullscreen
 - Engine
 - Ambient light
 - background
 - camera pose

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))

**Note to maintainers**: Remember to use **Squash-Merge**

Issue #137, opened 2020-10-06
Consolidate Scene3D with ign-gazebo's GzScene3D


chapulina:
We currently have 2 separate scene plugins in the Ignition ecosystem, one in `ign-gui` and one in `ign-gazebo`. The split into 2 scene plugins was something we did to move forward quickly, but the long-term goal is to combine those into a single plugin in a way that all Gazebo-independent features can be pushed into `ign-gui` so all downstream projects can take advantage of it and there's less duplication to maintain.

Some possible ways to achieve this:

* Move all Gazebo-specific features to separate plugins that interact with the 3D scene
* Make the `ign-gui` plugin extensible, similar to how we have [ignition::gui::TransportPlotting](https://github.com/ignitionrobotics/ign-gui/tree/main/src/plugins/plotting) and [ignition::gazebo::Plotting](https://github.com/ignitionrobotics/ign-gazebo/tree/main/src/gui/plugins/plotting).

scpeters:
we are using the `Scene3D` plugin in the delphyne project and have noticed some errors when loading mesh materials, but I believe this has already been fixed in ign-gazebo, so I'm cross-porting the fix from ign-gazebo's SceneManager in #191

ahcorde:
Related PRs

- MoveToHelper (ign-rendering) https://github.com/ignitionrobotics/ign-rendering/pull/311
- New Scene3D interface https://github.com/ignitionrobotics/ign-gui/pull/217
- New Scene3D plugin using the interface provided in ign-gui https://github.com/ignitionrobotics/ign-gazebo/pull/798

ahcorde:
@chapulina, I have been working in adding small plugins to ign-gazebo and ign-gui. I updated the ign-gazebo PR with some of the plugins.

 - ign-gui
   - view_control  https://github.com/ignitionrobotics/ign-gui/pull/231/
   - Move to / follow  https://github.com/ignitionrobotics/ign-gui/pull/226
 
- ign-gazebo
  - Select https://github.com/ignitionrobotics/ign-gazebo/pull/854
  - Transform controls https://github.com/ignitionrobotics/ign-gazebo/pull/854

ahcorde:
I have created some branches to test this (Fortress). Because it requires some cherry picks:

  - ign-gazebo https://github.com/ignitionrobotics/ign-gazebo/tree/ahcorde/6/scene_manager_test
[...]

PR #221, opened 2021-05-08
Split transport scene manager into a plugin outside Scene3D

chapulina:
# 🎉 New feature

Part of #137 , alternative to #217 

Needs https://github.com/ignitionrobotics/ign-rendering/pull/320

## Summary

This is a step towards reducing duplication between 3D scenes. Initially, @ahcorde and I had talked about taking an approach similar to the one used by plotting. That is providing an interface that is wrapped by an `ign-gui` plugin, and also wrapped by an `ign-gazebo` plugin, which adds Gazebo-specific functionality on top. @ahcorde started working on this approach in #217, and I also iterated a bit on top of it on 2fc70f28e91368043e87dceea72656256c96cbf2. But while at it, I realized that this approach may not be the best fit for `Scene3D`, because there's actually a simpler way.

My new idea is to leverage the `Render` event to perform all scene update tasks such as adding, removing and updating visuals. The `Scene3D` plugin itself should be very minimal, just focusing on running the render loop and providing some orbit controls. I'm leaving this minimal functionality on the new `MinimalScene` plugin, while all other functionality should be pushed to new plugins, like the `TransportSceneManager` that's being proposed here.

This approach allows us to keep Ignition GUI's `Scene3D` almost the same and add all new functionality on top. I think we could do the same with `ignition::gazebo::Scene3D`, see https://github.com/ignitionrobotics/ign-gazebo/pull/813/. That is:

* copy all the logic that is not already in `ignition::gui::Scene3D` to a new plugin, `GazeboSceneManager`
* `GazeboSceneManager` will:
    * not work by itself. Like `TransportSceneManager`, it must be loaded alongside `ignition::gui::MinimalScene`
    * have no code that paints to the window (like texture IDs, instantiating a rendering thread, etc). That's the `gui::MinimalScene`'s responsibility.
    * install a filter to the `Render` event and perform all rendering calls within that callback.
* Because we've been actively developing it, `gazebo::Scene3D` may have some functions and bugfixes that we'll need to port to `gui::MinimalScene`. We just need to be careful to push too much functionality into `gui::MinimalScene`. Let's keep that plugin lean. Any well scoped feature should go into a separate plugin that ties into the render event.

## Test it

Try the new `scene_provider` example, see instructions on README added.

## Checklist
- [ ] Added tests


chapulina:
## Updates

1. Moved the new approach to a new plugin, `MinimalScene`, so we can tick-tock the change for users. `Scene3D` is deprecated on v6 and will be removed on v7.
2. Improved the example so the box moves:

random_box

TODO:

- [ ] Add tests

chapulina:
@ahcorde , I'm opening this PR for review. I've added tests, a tutorial, and a migration path.

test/integration/minimal_scene.cc
```diff
@@ -0,0 +1,128 @@
...
+  ASSERT_NE(nullptr, root);
+  EXPECT_EQ(1u, root->ChildCount());
+
+  // Check camera
+  auto camera = std::dynamic_pointer_cast<rendering::Camera>(
+      root->ChildByIndex(0));
+  ASSERT_NE(nullptr, camera);
+
+  EXPECT_EQ(math::Pose3d(1, 2, 3, 0, 0, 1.57), camera->WorldPose());
+}
```
ahcorde, line 127: cleanup ? 

```cpp
 // Cleanup
  auto plugins = win->findChildren<Plugin *>();
  for (auto & p : plugins)  {
    auto pluginName = p->CardItem()->objectName();
    app.RemovePlugin(pluginName.toStdString());
  }
  win->QuickWindow()->close(); 
  engine->DestroyScene(scene);
  rendering::unloadEngine(engine->Name());
```

test/integration/transport_scene_manager.cc
```diff
@@ -0,0 +1,233 @@
...
+
+  sleep = 0;
+  while (root->ChildCount() > 1 && sleep < maxSleep)
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    QCoreApplication::processEvents();
+    sleep++;
+  }
+  EXPECT_EQ(1u, root->ChildCount());
+}
```
ahcorde, line 232: cleanup?

ahcorde (changes requested):

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -0,0 +1,796 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
chapulina: These files were mostly copied from `Scene3D`, so I kept the original copyright year

src/plugins/minimal_scene/MinimalScene.hh
```diff
@@ -0,0 +1,342 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
chapulina: Same here, keeping the copyright because the file is mainly a copy

test/integration/minimal_scene.cc
```diff
@@ -0,0 +1,142 @@
...
+
+using namespace ignition;
+using namespace gui;
+
+/////////////////////////////////////////////////
+TEST(MinimalSceneTest, IGN_UTILS_TEST_ENABLED_ONLY_ON_LINUX(Load))
+{
+  common::Console::SetVerbosity(4);
+
+  Application app(g_argc, g_argv);
```
ahcorde, line 47: Bionic is failing in this line, we should initialize the char**

test/integration/minimal_scene.cc
```diff
@@ -0,0 +1,142 @@
...
+
+#include "test_config.h"  // NOLINT(build/include)
+#include "../helpers/TestHelper.hh"
+#include "ignition/gui/Application.hh"
+#include "ignition/gui/GuiEvents.hh"
+#include "ignition/gui/Plugin.hh"
+#include "ignition/gui/MainWindow.hh"
+
+int g_argc = 1;
+char **g_argv = new char *[g_argc];
```
ahcorde, line 37:
```suggestion
char **g_argv = {"minimal_test"};
```

ahcorde, line 37: The first element should be initialize to avoid the crash on Bionic

ahcorde (changes requested):

test/integration/minimal_scene.cc
chapulina: I'm on Bionic and the test passes for me. It also passes on [GH actions](https://github.com/ignitionrobotics/ign-gui/pull/221/checks?check_run_id=2913506543). Does it fail locally for you?

The proposed change won't compile. This does:

```suggestion
char *g_argv[] = {"unused"};
```

But it triggers a warning:

```
warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]
```

I think we could suppress that warning, but I'm not sure that the change is needed at all.

ahcorde: I created a PR here https://github.com/ignitionrobotics/ign-gui/pull/244

According with the official [Qt documentation](https://doc.qt.io/qt-5/qguiapplication.html#QGuiApplication).

```
Warning: The data referred to by argc and argv must stay valid for the entire lifetime of the QGuiApplication object.
In addition, argc must be greater than zero and argv must contain at least one valid character string.
```

src/plugins/minimal_scene/MinimalScene.hh
```diff
@@ -0,0 +1,387 @@
...
+    // Documentation inherited
+    protected: bool eventFilter(QObject *_obj, QEvent *_event) override;
+
+    // Documentation inherited
+    public: virtual void LoadConfig(const tinyxml2::XMLElement *_pluginElem)
+        override;
+
+    /// \internal
+    /// \brief Pointer to private data.
+    private: std::unique_ptr<MinimalScenePrivate> dataPtr;
```
ahcorde, line 86: IGN_UTILS_IMPL_PTR ?

src/plugins/minimal_scene/MinimalScene.hh
```diff
@@ -0,0 +1,387 @@
...
+    /// New scene messages will be published to this topic when an entities are
+    /// added
+    public: std::string sceneTopic;
+
+    /// \brief True if sky is enabled;
+    public: bool skyEnable = false;
+
+    /// \internal
+    /// \brief Pointer to private data.
+    private: std::unique_ptr<IgnRendererPrivate> dataPtr;
```
ahcorde, line 201: IGN_UTILS_IMPL_PTR ?

src/plugins/minimal_scene/MinimalScene.hh
```diff
@@ -0,0 +1,387 @@
...
+    /// \param[in] _oldNode The node passed in previous updatePaintNode
+    /// function. It represents the visual representation of the item.
+    /// \param[in] _data The node transformation data.
+    /// \return Updated node.
+    private: QSGNode *updatePaintNode(QSGNode *_oldNode,
+        QQuickItem::UpdatePaintNodeData *_data) override;
+
+    /// \internal
+    /// \brief Pointer to private data.
+    private: std::unique_ptr<RenderWindowItemPrivate> dataPtr;
```
ahcorde, line 335: IGN_UTILS_IMPL_PTR

ahcorde (changes requested):

ahcorde (approved):

ahcorde:
@chapulina, these two PRs are the following ones:

 - https://github.com/ignitionrobotics/ign-gui/pull/231
 - https://github.com/ignitionrobotics/ign-gui/pull/226

PR #225, opened 2021-05-26
Confirmation dialog when closing main window

francocipollone:
Signed-off-by: Franco Cipollone <franco.c@ekumenlabs.com>

# 🎉 New feature

Related to https://github.com/ignitionrobotics/ign-gazebo/issues/820

## Summary

A new property (`dialog_on_exit`) was added to the `WindowConfig` struct to enable displaying a confirmation dialog when exiting the window.
In the `.config` file:
```xml
<!-- Window -->
<window>
  <width>1000</width>
  <height>845</height>
  ...
  ...
  <dialog_on_exit>true</dialog_on_exit>    // New Feature
</window>
```
 - Using this property in gazebo: [ign-gazebo branch](https://github.com/ignitionrobotics/ign-gazebo/commit/606a17b6eaac8296a1ee85bef4805f7bfdfcd83b)
 
 

https://user-images.githubusercontent.com/53065142/119691095-a87e1c00-be20-11eb-8128-7812714e1f3d.mp4

## Test it
```
// bring this changes.
cd ign-gui
git checkout francocipollone/citadel_confirmation_dialog
cd ../
// Enable this feature for gazebo app.
// Just one line of code.
cd ign-gazebo
git checkout francocipollone/citadel_confirmation_dialog
// Build and run gazebo.
colcon build --packages-select ign-gui3 ign-gazebo3 
ign gazebo
```
Press the X button or Quit from the menu.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/gui/MainWindow.hh
+++ b/include/ignition/gui/MainWindow.hh
@@ -44,7 +44,7 @@ namespace ignition
 
     /// \brief The main window class creates a QQuickWindow and acts as an
     /// interface which provides properties and functions which can be called
-    /// from MainWindow.qml
+    /// from Main.qml
     class IGNITION_GUI_VISIBLE MainWindow : public QObject
     {
       Q_OBJECT
@@ -175,6 +175,14 @@ namespace ignition
         READ ShowPluginMenu
         WRITE SetShowPluginMenu
         NOTIFY ShowPluginMenuChanged
+      )
+
+      /// \brief Flag to enable confirmation dialog on exit
+      Q_PROPERTY(
+        bool showDialogOnExit
+        READ ShowDialogOnExit
+        WRITE SetShowDialogOnExit
+        NOTIFY ShowDialogOnExitChanged
       )
 
       /// \brief Constructor
@@ -344,6 +352,14 @@ namespace ignition
       /// \param[in] _showPluginMenu True to show.
       public: Q_INVOKABLE void SetShowPluginMenu(const bool _showPluginMenu);
 
+      /// \brief Get the flag to show the plugin menu.
+      /// \return True to show.
+      public: Q_INVOKABLE bool ShowDialogOnExit() const;
+
+      /// \brief Set the flag to show the confirmation dialog when exiting.
+      /// \param[in] _showDialogOnExit True to show.
+      public: Q_INVOKABLE void SetShowDialogOnExit(bool _showDialogOnExit);
+
       /// \brief Callback when load configuration is selected
       public slots: void OnLoadConfig(const QString &_path);
 
@@ -397,6 +413,9 @@ namespace ignition
 
       /// \brief Notifies when the show menu flag has changed.
       signals: void ShowPluginMenuChanged();
+
+      /// \brief Notifies when the showDialogOnExit flag has changed.
+      signals: void ShowDialogOnExitChanged();
 
       /// \brief Notifies when the window config has changed.
       signals: void configChanged();
@@ -483,6 +502,9 @@ namespace ignition
 
       /// \brief Show the plugins menu
       bool showPluginMenu{true};
+
+      /// \brief Show the confirmation dialog on exit
+      bool showDialogOnExit{false};
 
       /// \brief True if plugins found in plugin paths should be listed under
       /// the Plugins menu. True by default.
--- a/include/ignition/gui/MainWindow.hh
+++ b/include/ignition/gui/MainWindow.hh
@@ -503,9 +503,6 @@ namespace ignition
       /// \brief Show the plugins menu
       bool showPluginMenu{true};
 
-      /// \brief Show the confirmation dialog on exit
-      bool showDialogOnExit{false};
-
       /// \brief True if plugins found in plugin paths should be listed under
       /// the Plugins menu. True by default.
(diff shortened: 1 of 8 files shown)


include/ignition/gui/MainWindow.hh
```diff
@@ -484,6 +503,9 @@ namespace ignition
       /// \brief Show the plugins menu
       bool showPluginMenu{true};
 
+      /// \brief Show the confirmation dialog on exit
+      bool showDialogOnExit{false};
```
jennuine, line 507: New publicly visible data members break ABI, to avoid this you can add it to the [`MainWindowPrivate` class](https://github.com/ignitionrobotics/ign-gui/blob/890a5722640c539068b4c658a8723a74930a9cd6/src/MainWindow.cc#L33) in `MainWindow.cc` and access it using `this->dataPtr`

src/MainWindow.cc
```diff
@@ -470,6 +477,12 @@ bool WindowConfig::MergeFromXML(const std::string &_windowXml)
     }
   }
 
+  // Show dialog on exit
+  if (auto dialogOnExitElem = winElem->FirstChildElement("dialog_on_exit"))
```
jennuine, line 481: In my opinion, I think the confirmation dialog should always appear by default (regardless if `<dialog_on_exit>true</dialog_on_exit>` is in `gui.config` or not) and when the user does not want the confirmation then they would set `<dialog_on_exit>false</dialog_on_exit>` but lets see what others think

cc @mabelzhang @ahcorde

jennuine (changes requested):
Looks good! I left a couple comments

src/MainWindow.cc
mabelzhang: I don't have a strong opinion either way. I think it is safer for it to be on by default, but then I checked Gazebo-classic, and it doesn't seem to have a confirmation dialog. So... :woman_shrugging:

ahcorde: In general I would to stick with the current behaviour and document this new feature. I open and close many times ign-gazebo and this is adding an overhead for me.

Maybe we can discuss in the next ignition meeting

include/ignition/gui/MainWindow.hh
francocipollone: Oh my bad, you are right. Thanks for the suggestion.

Check 50ac123:
In order to avoid breaking ABI I added that member variable to `MainWindowPrivate` class and it was removed from `WindowsConfig`. This change made me move one level up the parsing of the `dialog_on_exit` XML node (to `Application.cc`). Besides, the `dialog_on_exit` XML attribute isn't part of `window` XML node anymore.
It ended up not being as tidy as before. PTAL

src/MainWindow.cc
francocipollone: I think that leaving this off by default is the safer option. 
Although `ignition-gazebo` is the direct consumer of `ignition-gui` we probably don't want to change the behavior of all the applications that are developed on top of `ignition::gui::Application` (extern projects).

src/Application.cc
```diff
@@ -275,6 +275,14 @@ bool Application::LoadConfig(const std::string &_config)
     this->dataPtr->windowConfig.MergeFromXML(std::string(printer.CStr()));
   }
 
+  // Closing behavior.
+  if (auto dialogOnExitElem = doc.FirstChildElement("dialog_on_exit"))
+  {
+    bool showDialogOnExit{false};
+    dialogOnExitElem->QueryBoolText(&showDialogOnExit);
+    this->dataPtr->mainWin->SetShowDialogOnExit(showDialogOnExit);
+  }
```
jennuine, line 284: I think the previous behavior where `<dialog_on_exit>` was in `<window>` was better. You can move this section to the above `if (auto winElem = doc.FirstChildElement("window"))` and then update this `if` to `if (auto dialogOnExitElem = winElem->FirstChildElement("dialog_on_exit"))`

jennuine (commented):
Looks good, left a minor comment. Also, there are a couple of tests failing. `UNIT_Application_TEST` might be flaky but `UNIT_MainWindow_TEST` is seg faulting at `MainWindowTest.Constructor`

src/Application.cc
francocipollone: Agree, from a user perspective it is better to have it within the `<window>` node.
Done!

jennuine (commented):
Thanks for iterating! Can you add a test for codecov? Adding `<window>` and `<dialog_on_exit>`  to `test/config/test.config` should suffice. Also, add the new `<dialog_on_exit>` element under "Overview" to [layout tutorials](https://github.com/ignitionrobotics/ign-gui/blob/ign-gui3/tutorials/04_layout.md)

francocipollone:
> Thanks for iterating! Can you add a test for codecov? Adding `<window>` and `<dialog_on_exit>` to `test/config/test.config` should suffice. Also, add the new `<dialog_on_exit>` element under "Overview" to [layout tutorials](https://github.com/ignitionrobotics/ign-gui/blob/ign-gui3/tutorials/04_layout.md)

Thanks for the quick re-review!
Addressed! Check ceaaedc and 6db723c

jennuine (approved):

scpeters (approved):

PR #228, opened 2021-05-28
Added Mouse Events

ahcorde:

# 🎉 New feature

Related with this issue https://github.com/ignitionrobotics/ign-gui/issues/209. I added left and right mouse events. These events are generated by the Scene3D window, otherwise if we try to get the mouse click from other plugin we will get the click in the hole window making more difficult to handle it. Other advantage it's that we are publishing a ignition related event.

## Checklist
- [x] Added tests


mjcarroll (approved):
LGTM, but the `INTEGRATION_Scene3d` test is segfaulting now.

mjcarroll:

include/ignition/gui/GuiEvents.hh
```diff
@@ -255,6 +259,59 @@ namespace ignition
...
+        /// \param[in] _mouse The left mouse event
+        /// the scene
+        public: explicit LeftClickOnScene(
+          const ignition::common::MouseEvent &_mouse)
+            : QEvent(kType), mouse(_mouse)
+        {
+        }
+
+        /// \brief Unique type for this event.
+        static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser - 10);
```
chapulina, line 277: Why skip `-8` and `-9`?

chapulina (commented):
Have you looked into the `INTEGRATION_Scene3d` test failure?

include/ignition/gui/GuiEvents.hh
chapulina: Oh I see them in https://github.com/ignitionrobotics/ign-gui/pull/224

chapulina (changes requested):
Do you have a PR that uses these events, @ahcorde ? It would be nice to make sure the events have everything we need before merging and releasing. These classes don't use the PIMPL pattern, so if we realize later that they need more fields, we can't break ABI. See https://github.com/ignitionrobotics/ign-gui/issues/209#issuecomment-849808255 for more context.

ahcorde:
@chapulina this PR make use of it https://github.com/ignitionrobotics/ign-gui/pull/231/

ahcorde:
friendly ping @chapulina

chapulina:
@ahcorde , how about we target this one at `main` so we can iterate on it a bit. Then if we see the need to backport we can. It doesn't look like we have a use case for this on Citadel yet, right?

ahcorde:
I will wait to this forward port https://github.com/ignitionrobotics/ign-gui/pull/238 because it includes the scene3d tests.

ahcorde:
@chapulina retargeted to `main` and add PIMPL pattern

chapulina (approved):
I pushed some documentation and tweaks in edcfb77be72a9f25b7edd4f5043c4e1d7d873fc4.

I'm on the fence about keeping both the `ClickOnScene` and `ClickToScene` events. I think they could be combined into a single event to avoid confusion. But I'll refrain from making more changes until @scpeters opens a PR adding the distance field as proposed in https://github.com/ignitionrobotics/ign-gui/issues/209#issuecomment-849808255.

Let's go with these for now!

PR #226, opened 2021-05-26
Added camera tracking

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

# 🎉 New feature

## Summary

This PR is part of the consolidation between the scene3d in ign-gui and ign-gazebo.

This plugin allows to `follow` and `moveto` the camera. This requires to change a small detail in ign-gui scene3D when we create the camera, instead of using the method `createCamera(void)` we should use `createCamera(string)`. Then we will be able to fetch the camera from this plugin.

Related PRs:
 - https://github.com/ignitionrobotics/ign-gui/pull/221
 - https://github.com/ignitionrobotics/ign-gazebo/pull/813
 - https://github.com/ignitionrobotics/ign-rendering/pull/320#event-4749407284
 - Emit more events https://github.com/ignitionrobotics/ign-gui/pull/213 and https://github.com/ignitionrobotics/ign-gui/pull/224
 - MoveTo in ign-rendering https://github.com/ignitionrobotics/ign-rendering/pull/311

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #231, opened 2021-06-08
Added Interactive view control plugin

ahcorde:

# 🎉 New feature

## Summary

This PR is part of the consolidation between the scene3d in ign-gui and ign-gazebo.

This plugin allows to control the camera in the 3D Scene.

Related PRs:
 - https://github.com/ignitionrobotics/ign-gui/pull/221
 - https://github.com/ignitionrobotics/ign-gazebo/pull/813
 - https://github.com/ignitionrobotics/ign-rendering/pull/320#event-4749407284
 - Emit more events https://github.com/ignitionrobotics/ign-gui/pull/213 and https://github.com/ignitionrobotics/ign-gui/pull/224
 - MoveTo in ign-rendering https://github.com/ignitionrobotics/ign-rendering/pull/311

## Checklist
- [ ] Added tests


src/plugins/view_control/ViewControl.hh
```diff
@@ -0,0 +1,59 @@
...
+
+namespace ignition
+{
+namespace gui
+{
+namespace plugins
+{
+  class ViewControlPrivate;
+
+  /// \brief
```
chapulina, line 33: Missing docs

src/plugins/scene3d/Scene3D.cc
```diff
@@ -210,18 +209,12 @@ namespace plugins
     /// \brief Mouse event
     public: common::MouseEvent mouseEvent;
 
-    /// \brief Mouse move distance since last event.
```
chapulina, line 213: How about we leave `Scene3D` as it is and focus on `MinimalScene`? So Edifice users can migrate to Fortress and the scene still works without any changes, but then they have the option of migrating to `MinimalScene` + other plugins.

src/plugins/view_control/ViewControl.cc
```diff
@@ -0,0 +1,241 @@
...
+void ViewControlPrivate::OnRender()
+{
+  if (!this->scene)
+  {
+    this->scene = rendering::sceneFromFirstRenderEngine();
+    if (!this->scene)
+      return;
+
+    this->camera = std::dynamic_pointer_cast<rendering::Camera>(
+      this->scene->SensorByName("Scene3DCamera"));
```
chapulina, line 83: I think that hardcoding the name of the camera may be a bit limiting. Can we assume that the first camera is the user camera? Like the screenshot plugin does:

https://github.com/ignitionrobotics/ign-gui/blob/d509ab956e50527daa6005345606b77c4f980bbf/src/plugins/screenshot/Screenshot.cc#L187-L198

src/plugins/view_control/ViewControl.cc
```diff
@@ -0,0 +1,241 @@
...
+    if (!this->camera)
+    {
+      ignerr << "TransformControlLogic Camera is not available" << std::endl;
```
chapulina, line 86:
```suggestion
      ignerr << "ViewControl camera is not available" << std::endl;
```

include/ignition/gui/GuiEvents.hh
```diff
@@ -255,6 +255,24 @@ namespace ignition
         /// for this event.
         private: bool menuEnabled;
       };
+
+      class BlockOrbit : public QEvent
```
chapulina, line 259: What's the intended use case for this event? I don't see it being used.

chapulina (commented):
Can we add this plugin to `scene3d.config`?

> This plugin allows to control the camera in the 3D Scene.

It's worth remembering that there are multiple ways of controlling the user camera, and they're being spread across various plugins:

* `ViewAngle` provides canonical angles, set precise pose and projection type
* The plugin added in #226 provides "follow" and "move to"
* This plugin allows orbiting with a mouse

In the future we may want to have more camera control functionality:

* First person view control with mouse (I think it would be part of this plugin)
* Keyboard based view control (probably also part of this plugin)

I can imagine this being confusing for users in the future.

I think this plugin is basic enough that it could be part of `MinimalScene`. On the other hand, [some users](https://answers.gazebosim.org//question/8694/mouse-control-buttons/) may want to implement different orbit controls, so we either make this configurable, or let them implement their own thing. So since you've already started moving this to a new plugin, let's keep this separated.

I just think we should work on the names for all these plugins, so it's clear what each of them is for. This plugin is for orbiting with the mouse, possibly adding FPS control with the keyboard in the future. How about calling it `InteractiveViewControl`?

chapulina:
I should mention here that I updated https://github.com/ignitionrobotics/ign-gui/pull/221 when https://github.com/ignitionrobotics/ign-gui/pull/213 was merged forward. This is what's causing the conflicts here.

include/ignition/gui/GuiEvents.hh
ahcorde: BlockOrbit is emitted when the transform mode is on. it used here https://github.com/ignitionrobotics/ign-gazebo/pull/854/files

src/plugins/scene3d/Scene3D.cc
ahcorde: https://github.com/ignitionrobotics/ign-gui/pull/231/commits/0f4954cc1c4e87234de823950938375aad3875d3

chapulina:
> How about calling it InteractiveViewControl?

Did you give this any thought, @ahcorde ?

ahcorde:
> > How about calling it InteractiveViewControl?
> 
> Did you give this any thought, @ahcorde ?

Sorry I didn't see that comment.  Done https://github.com/ignitionrobotics/ign-gui/pull/231/commits/f3de5463e94e01ea9bebf640d1034be7356e6898

src/plugins/view_control/ViewControl.cc
ahcorde: With the recent changes in the PR of running in the process this may be not true. I included this change, I think all cameras except the one is moving should have a name similar to `scene::camera(XXXX)` otherwise should be named like `model_name::link_name::camera_name`

https://github.com/ignitionrobotics/ign-gui/blob/f3de5463e94e01ea9bebf640d1034be7356e6898/src/plugins/interactive_view_control/InteractiveViewControl.cc#L82-L95

chapulina: > With the recent changes in the PR of running in the process this may be not true

Yeah I see your point. The problem is that the scene of `Scene3D` or `MinimalScene` may not be called `scene` if the user passes a custom name through XML's `<scene>` element. I ran into this while writing a test for #226 which had a different scene name.

So I think we need a different way of identifying the user camera. On Gazebo classic there was a separate class for the user camera, so that was easy. On Ignition Rendering I don't think there's any distinction.

Another thing to keep in mind is that in the future we may want to restore support for multiple user cameras in the same scene, and these plugins that attach to a camera may want to choose which camera to attach to ([like we had in `ign-gui0`](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gui/pull-requests/74/page/1)).

Once https://github.com/ignitionrobotics/ign-rendering/issues/58 is tackled, maybe we could store a different property into the user camera(s)? @iche033 , do you have other ideas on how to identify if a camera is on the GUI or a sensor?

ahcorde: PR https://github.com/ignitionrobotics/ign-rendering/pull/358

src/plugins/interactive_view_control/InteractiveViewControl.cc
```diff
@@ -0,0 +1,259 @@
...
+  public: math::Vector2d drag;
+
+  /// \brief User camera
+  public: rendering::CameraPtr camera{nullptr};
+
+  /// \brief View control focus target
+  public: math::Vector3d target;
+
+  /// \brief Camera orbit controller
+  public: rendering::OrbitViewController viewControl;
```
chapulina, line 66: Note that we'll also need to support ortho control, added in https://github.com/ignitionrobotics/ign-gazebo/pull/815.

That's where I think that this plugin starts overlapping with `ViewAngle`. I'm not sure if we should put the ortho logic here, but keep the button on `ViewAngle`. I guess we can start with that and see if it becomes too confusing for users?

include/ignition/gui/GuiEvents.hh
```diff
@@ -351,6 +351,20 @@ namespace ignition
         /// \brief Private data pointer
         IGN_UTILS_IMPL_PTR(dataPtr)
       };
+
+      class IGNITION_GUI_VISIBLE BlockOrbit : public QEvent
```
chapulina, line 355: Please document how this event should be used. Also document all functions.

chapulina (commented):
I did a first pass and pushed some trivial changes in 7dce9b0ddbbe03317fac1184a955e99ddf7e85f2.

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -892,7 +838,8 @@ void RenderWindowItem::wheelEvent(QWheelEvent *_e)
 #endif
   double scroll = (_e->angleDelta().y() > 0) ? -1.0 : 1.0;
   this->dataPtr->renderThread->ignRenderer.NewMouseEvent(
-      this->dataPtr->mouseEvent, math::Vector2d(scroll, scroll));
+    this->dataPtr->mouseEvent);
+  this->dataPtr->mouseEvent.SetScroll(scroll, scroll);
```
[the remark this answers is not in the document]
ahcorde: Not really sure about this... uhmm

src/plugins/interactive_view_control/InteractiveViewControl.cc
```diff
@@ -0,0 +1,259 @@
...
+
+#include "InteractiveViewControl.hh"
+
+/// \brief Private data class for InteractiveViewControl
+class ignition::gui::plugins::InteractiveViewControlPrivate
+{
+  /// \brief
+  public: void OnRender();
+
+  math::Vector3d ScreenToScene(
```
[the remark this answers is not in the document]
ahcorde: where do you think this should live? ign-rendering ?

ahcorde:
Requires https://github.com/ignitionrobotics/ign-rendering/pull/358

src/plugins/interactive_view_control/InteractiveViewControl.cc
ahcorde: Added ortho control https://github.com/ignitionrobotics/ign-gui/pull/231/commits/c81302dd505d90d69fc872fc722b8de322ccd497

ahcorde:

ahcorde:
cleaned commit history because this PR was targeting a different branch

ahcorde:
Windows warnings should be fixed with this PR https://github.com/ignitionrobotics/ign-gui/pull/254

chapulina (approved):
LGTM! I pushed some docs and small tweaks in 323471f902d3b1c8cf988371178f59bd640f9ec8. Feel free to merge if those look ok to you, @ahcorde .

PR #235, opened 2021-06-20
Fix crash when starting Grid3D plugin

darksylinc:
If `Grid3DPrivate::engine` happens not to be a nullptr during
allocation, it will never be correctly initialized and will later try to
dereference an invalid ptr

Just in case we also initialize quickWindow although it doesn't seem to
be strictly necessary.

Signed-off-by: Matias N. Goldberg <dark_sylinc@yahoo.com.ar>

## Summary

Self-explanatory. I ran into this bug by chance. Just open the Grid3D plugin from the GUI and it will crash (unless you're lucky that Grid3DPrivate::engine ended up being nullptr).

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/grid_3d/Grid3D.cc
+++ b/src/plugins/grid_3d/Grid3D.cc
@@ -88,11 +88,11 @@ namespace plugins
   class Grid3DPrivate
   {
     /// brief Parent window
-    public: QQuickWindow *quickWindow;
+    public: QQuickWindow *quickWindow{nullptr};
 
     /// \brief We keep a pointer to the engine and rely on it not being
     /// destroyed, since it is a singleton.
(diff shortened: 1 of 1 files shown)


ahcorde (commented):
This is already fixed [here](https://github.com/ignitionrobotics/ign-gui/commit/bf50949a2634decdb014792bde1378d71b61c454). We need to forward port the commit.

chapulina:
Thanks for the PR! As @ahcorde mentioned, we just need to forward-port the fix, I'll decline this in favor of that. 

The 3 :arrow_right: 4 port is in https://github.com/ignitionrobotics/ign-gui/pull/234, then we can open a 4 :arrow_right: 5.

chapulina:
Done in #238

PR #240, opened 2021-06-23
Detect ign instead of using cmake module to check for ignition-tools

j-rivero:
Part of ignition-tooling/release-tools#472

I don't detect the conditional installation of the `ign` submodule, the PR just
changed the way of detecting ign for testing.


jennuine (approved):
LGTM

chapulina (approved):

PR #247, opened 2021-07-01
Added MarkerManager Plugin 

ahcorde:
# 🎉 New feature

## Summary

This PR is part of the consolidation between the Scene3D in ign-gui and ign-gazebo. This PR add a marker manager plugin

## Test it

 - Compile the example `marker` inside the repo `ign-gazebo`
 - Add this lines to your `gui.config` file
```xml
<plugin filename="MarkerManager" name="Marker manager">
  <ignition-gui>
    <title>Marker Manager</title>
    <property type="bool" key="showTitleBar">false</property>
    <property type="string" key="state">docked</property>
    <property key="resizable" type="bool">false</property>
    <property key="width" type="double">5</property>
    <property key="height" type="double">5</property>
    <property key="state" type="string">floating</property>
    <property key="showTitleBar" type="bool">false</property>
  </ignition-gui>
</plugin>
```
 - launch gazebo `ign gazebo --gui-config src/ign-gazebo/src/gui/gui.config -v 4`
 - Run the `marker` example

## Checklist
- [ ] Added tests


src/plugins/marker_manager/MarkerManager.hh
```diff
@@ -0,0 +1,57 @@
...
+
+namespace ignition
+{
+namespace gui
+{
+namespace plugins
+{
+  class MarkerManagerPrivate;
+
+  class MarkerManager : public Plugin
```
chapulina, line 33: Docs :wink:

src/plugins/marker_manager/MarkerManager.cc
```diff
@@ -0,0 +1,672 @@
...
+};
+
+using namespace ignition;
+using namespace gui;
+using namespace plugins;
+
+/////////////////////////////////////////////////
+void MarkerManagerPrivate::Initialize()
+{
+  igndbg << "Initializing Marker Manager plugin\n";
```
chapulina, line 136: This can be removed

src/plugins/marker_manager/MarkerManager.cc
```diff
@@ -0,0 +1,672 @@
...
+  if (this->title.empty())
+    this->title = "Marker Manager";
+
+  // Custom parameters
+  if (_pluginElem)
+  {
+    auto elem = _pluginElem->FirstChildElement("topic_name");
+    if (nullptr != elem && nullptr != elem->GetText())
+    {
+      this->dataPtr->topicName = elem->GetText();
```
chapulina, line 652: We could get this `AsValidTopic` to protect against bad user input

src/plugins/marker_manager/MarkerManager.cc
```diff
@@ -0,0 +1,672 @@
...
+  public: std::map<std::string,
+      std::map<uint64_t, ignition::rendering::VisualPtr>> visuals;
+
+  /// \brief Ignition node
+  public: ignition::transport::Node node;
+
+  /// \brief Topic name for the marker service
+  public: std::string topicName = "/marker";
+
+  /// \brief Sim time according to UpdateInfo in RenderUtil
```
chapulina, line 122: Update docs

src/plugins/marker_manager/MarkerManager.cc
```diff
@@ -0,0 +1,672 @@
...
+void MarkerManagerPrivate::OnWorldStatsMsg(
+  const ignition::msgs::WorldStatistics &_msg)
+{
+  std::chrono::steady_clock::duration timePoint;
+  if (_msg.has_sim_time())
+  {
+    timePoint = math::secNsecToDuration(
+        _msg.sim_time().sec(),
+        _msg.sim_time().nsec());
+    this->simTime = timePoint;
```
chapulina, line 616: Should we protect `simTime` with a mutex? The messages are received in a different thread from where the time is used

src/plugins/marker_manager/MarkerManager.cc
```diff
@@ -0,0 +1,672 @@
...
+  }
+
+  if (_msg.has_real_time())
```
chapulina, line 619: When the message has both sim and real time, the current logic will always use real time. I think we should default to sim time, because that's what users usually care about. If we want, we could add an option for users to configure whether they want to use real time.

chapulina (commented):
Works well for me :+1: 

How about moving the `marker.cc` example from `ign-gazebo` here? I also think we should add this plugin to `scene3d.config`.

ahcorde:
- moved `marker.cc` example
 - added plugin to `scene3d.config`
 - Added world statistic topic to scene provider

ahcorde:
Added MarkerManager_TEST

chapulina (commented):
Note that the test is segfaulting

chapulina (approved):
:rocket:

PR #272, opened 2021-08-15
Added log storing for ign-gui

NickNair:
# 🎉 New feature

Closes [#962](https://github.com/ignitionrobotics/ign-gazebo/issues/962) from ign-gazebo repository

## Summary
Every time a simulation is run with the GUI, the GUI console logs are saved to ~/.ignition/gazebo/log/<timestamp>/gui_console.log

## Test it
Run a simulation. Then :
`cd /.ignition/gazebo/log/<TIMESTAMP>

cat gui_console.log`

## Checklist
- [ ] Added tests

### DIFF
--- a/src/ign.cc
+++ b/src/ign.cc
@@ -34,22 +34,28 @@ char* g_argv[] =
 };
 
 //////////////////////////////////////////////////
-extern "C" IGNITION_GUI_VISIBLE char *ignitionVersion()
+void startConsoleLog()
 {
   std::string home;
   ignition::common::env(IGN_HOMEDIR, home);
 
-  std::string recordPathMod = ignition::common::joinPaths(home,
-        ".ignition", "gazebo", "log",ignition::common::timeToIso(IGN_SYSTEM_TIME()));
-  ignLogInit(recordPathMod, "gui_console.log");
+  std::string logPathMod = ignition::common::joinPaths(home,
+      ".ignition", "gui", "log",
+      ignition::common::timeToIso(IGN_SYSTEM_TIME()));
+  ignLogInit(logPathMod, "console.log");
+}
 
-
+//////////////////////////////////////////////////
+extern "C" IGNITION_GUI_VISIBLE char *ignitionVersion()
+{
   return strdup(IGNITION_GUI_VERSION_FULL);
 }
 
 //////////////////////////////////////////////////
 extern "C" IGNITION_GUI_VISIBLE void cmdPluginList()
 {
+  startConsoleLog();
+
   ignition::gui::Application app(g_argc, g_argv);
 
   auto pluginsList = app.PluginList();
@@ -73,6 +79,8 @@ extern "C" IGNITION_GUI_VISIBLE void cmdPluginList()
 //////////////////////////////////////////////////
 extern "C" IGNITION_GUI_VISIBLE void cmdStandalone(const char *_filename)
 {
+  startConsoleLog();
+
   ignition::gui::Application app(g_argc, g_argv,
       ignition::gui::WindowType::kDialog);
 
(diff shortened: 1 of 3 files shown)


NickNair:
How do I write a test for this PR?

src/ign.cc
```diff
@@ -36,6 +36,14 @@ char* g_argv[] =
 //////////////////////////////////////////////////
 extern "C" IGNITION_GUI_VISIBLE char *ignitionVersion()
 {
+  std::string home;
+  ignition::common::env(IGN_HOMEDIR, home);
+
+  std::string recordPathMod = ignition::common::joinPaths(home,
+        ".ignition", "gazebo", "log",ignition::common::timeToIso(IGN_SYSTEM_TIME()));
```
chapulina, line 43: Ignition GUI is used by other applications besides Gazebo, so we should save this to `~/.ignition/gui` by default.

```suggestion
        ".ignition", "gui", "log",ignition::common::timeToIso(IGN_SYSTEM_TIME()));
```

For `ign-gazebo`, we can add the same logic to its `ign.cc`. It shouldn't conflict with this one because the `ign gazebo` command doesn't directly call the `ign gui` command.

chapulina (commented):
Thank you for the PR! I just have a small comment. Nice idea to put this into `ignitionVersion`, since that's called every time the user runs `ign gui`!

chapulina:
> How do I write a test for this PR?

You could write a test similar to this one:

https://github.com/ignitionrobotics/ign-gui/blob/6cdc2f7021e9435d62560f68c273ffad2f6a19d3/src/ign_TEST.cc#L56-L61

And check that the file was created.

I recommend setting the `$HOME` environment variable to a fake path like this:

https://github.com/ignitionrobotics/ign-gazebo/blob/7d1a3e09dc596a4bbee968a955949b55b08629cd/test/helpers/EnvTestFixture.hh#L43-L44

So that when the user runs tests, no files are created in their home folder.

src/ign.cc
NickNair: Oh ok. So there's going to be gui logs in both `~/.ignition/gui` and `~/.ignition/gazebo` ?

chapulina: When users run `ign gui`, the logs will end up in `~/.ignition/gui`, when they run `ign gazebo`, they should go into `~/.ignition/gazebo`.

chapulina: Done in 
b9d8345

chapulina (approved):
Thanks again for the PR, @NickNair !

I rebased on top of `ign-gui3` (Citadel), and after merged we can merge it forward.

Also pushed these changes in b9d8345:

* Changed directory from `gazebo` to `gui`
* Moved functionality to a new function for clarity, so in case we stop calling `ignitionVersion` every time, this still works
* Added a test
* Added a note to the tutorial

Maybe @jennuine can take a look too since I've touched a lot of the PR?

src/ign_TEST.cc
```diff
@@ -52,10 +54,43 @@ std::string custom_exec_str(std::string _cmd)
...
+
+  /// \brief Store user's real $HOME to set it back at the end of tests.
+  public: std::string realHome;
+};
+
 // See https://github.com/ignitionrobotics/ign-gui/issues/75
-TEST(CmdLine, IGN_UTILS_TEST_ENABLED_ONLY_ON_LINUX(list))
+TEST_F(CmdLine, IGN_UTILS_TEST_ENABLED_ONLY_ON_LINUX(list))
 {
+  EXPECT_TRUE(common::removeAll(this->kFakeHome));
```
jennuine, line 87: The test is failing here, I'm assuming it's returning false because there is no directory to be removed. I see that we are setting the env variable to `kFakeHome` but I don't see where the directory is created. Is this needed or could we just call `common::removeAll` without the expectation?

jennuine (commented):
LGTM, one minor comment

src/ign_TEST.cc
chapulina: Ouch, fixed in 6737a83f24916e64772d3e2f75af9b86c2d905fa

jennuine (approved):

osrf-triage:

osrf-triage:

osrf-triage:

PR #273, opened 2021-08-18
Create new GuiEvent HoverOnScene

ahcorde:
# 🎉 New feature

## Summary

It publishes the onhover mouse 2D points

Required by: 
 - https://github.com/ignitionrobotics/ign-gazebo/pull/983

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/gui/GuiEvents.hh
+++ b/include/ignition/gui/GuiEvents.hh
@@ -323,6 +323,29 @@ namespace ignition
         /// \brief Private data pointer
         IGN_UTILS_IMPL_PTR(dataPtr)
       };
+
+      /// \brief Event which is called to broadcast the 2D coordinates of a
+      /// user's mouse hover within the scene.
+      class IGNITION_GUI_VISIBLE HoverOnScene : public QEvent
+      {
+        /// \brief Constructor
+        /// \param[in] _point The point at which the mouse is hovering within
+        /// the scene
+        public: explicit HoverOnScene(const math::Vector2i &_point);
+
+        /// \brief Unique type for this event.
+        static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser - 13);
+
+        /// \brief Get the point within the scene over which the user is
+        /// hovering.
+        /// \return The 2D point
+        public: math::Vector2i Point() const;
+
+        /// \internal
+        /// \brief Private data pointer
+        IGN_UTILS_IMPL_PTR(dataPtr)
+      };
+
     }
   }
 }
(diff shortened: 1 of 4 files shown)


include/ignition/gui/GuiEvents.hh
```diff
@@ -323,6 +323,29 @@ namespace ignition
...
       };
+
+      /// \brief Event which is called to broadcast the 2D coordinates of a
+      /// user's mouse hover within the scene.
+      class IGNITION_GUI_VISIBLE HoverOnScene : public QEvent
+      {
+        /// \brief Constructor
+        /// \param[in] _point The point at which the mouse is hovering within
+        /// the scene
+        public: explicit HoverOnScene(const math::Vector2i &_point);
```
chapulina, line 334: How about using `common::MouseEvent` like we're doing on `(Right|Left)ClickOnScene`?

chapulina (changes requested):

include/ignition/gui/GuiEvents.hh
ahcorde: @chapulina 

The hover event in the QML is just providing these two values

https://github.com/ignitionrobotics/ign-gui/blob/6f50c9c2212705f83a78591de9f5943a54eca11f/src/plugins/minimal_scene/MinimalScene.qml#L42-L44

Another option is to extend the current `common::MouseEvent` to allow hover events:

https://github.com/ignitionrobotics/ign-common/blob/491d95588d0f5bcea4d4b3d4ee08b7a2ef982cb2/events/include/ignition/common/MouseEvent.hh#L54-L70

or `MOVE` is the same as hover ?

chapulina: Yeah `MOVE` with false `Dragging` is a hover :wink:

chapulina: I think the main advantage of `MouseEvent` is that we can add modifier keys to the event later as needed. For example, hover + Shift

include/ignition/gui/GuiEvents.hh
```diff
@@ -323,6 +323,29 @@ namespace ignition
...
+
+        /// \brief Unique type for this event.
+        static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser - 13);
```
jennuine, line 337: Noob question, why `-13`? Can this be documented in the comments?

chapulina: Good idea to comment it, because it keeps coming up :sweat_smile: 

tl;dr: `gazebo::gui:GuiEvents` count up, `gui::GuiEvents` count down.

See context here: https://github.com/ignitionrobotics/ign-gui/pull/70#discussion_r436873992

chapulina (approved):

PR #284, opened 2021-09-16
Fixed Right click press position

ahcorde:

# 🦟 Bug fix

## Summary

The right click press position is not being published.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -248,6 +248,8 @@ void IgnRenderer::BroadcastRightClick()
 
   events::RightClickToScene rightClickToSceneEvent(pos);
   App()->sendEvent(App()->findChild<MainWindow *>(), &rightClickToSceneEvent);
+  events::RightClickOnScene rightClickOnSceneEvent(this->dataPtr->mouseEvent);
+  App()->sendEvent(App()->findChild<MainWindow *>(), &rightClickOnSceneEvent);
 }
 
 
@@ -841,6 +843,7 @@ void RenderWindowItem::keyReleaseEvent(QKeyEvent *_e)
 void RenderWindowItem::mouseReleaseEvent(QMouseEvent *_e)
 {
   this->dataPtr->mouseEvent = convert(*_e);
+  this->dataPtr->mouseEvent.SetPressPos(_e->pos().x(), _e->pos().y());
 
   this->dataPtr->renderThread->ignRenderer.NewMouseEvent(
       this->dataPtr->mouseEvent);
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -1044,9 +1044,9 @@ void RenderWindowItem::OnDropped(const QString &_drop,
 /////////////////////////////////////////////////
 void RenderWindowItem::mousePressEvent(QMouseEvent *_e)
 {
-  auto event = convert(*_e);
-  event.SetPressPos(event.Pos());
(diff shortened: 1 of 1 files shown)


src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -248,6 +248,8 @@ void IgnRenderer::BroadcastRightClick()
 
   events::RightClickToScene rightClickToSceneEvent(pos);
   App()->sendEvent(App()->findChild<MainWindow *>(), &rightClickToSceneEvent);
+  events::RightClickOnScene rightClickOnSceneEvent(this->dataPtr->mouseEvent);
+  App()->sendEvent(App()->findChild<MainWindow *>(), &rightClickOnSceneEvent);
```
chapulina, line 252: Should we do the same for the left click?

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -841,6 +843,7 @@ void RenderWindowItem::keyReleaseEvent(QKeyEvent *_e)
 void RenderWindowItem::mouseReleaseEvent(QMouseEvent *_e)
 {
   this->dataPtr->mouseEvent = convert(*_e);
+  this->dataPtr->mouseEvent.SetPressPos(_e->pos().x(), _e->pos().y());
```
chapulina, line 846: This event is `mouseRelease`, not `mousePress`, so I think this may be misplaced

src/plugins/minimal_scene/MinimalScene.cc
ahcorde: it's done here

https://github.com/ignitionrobotics/ign-gui/blob/bde1977b4600602fb63f08d74cfbeeaf5bd3919b/src/plugins/minimal_scene/MinimalScene.cc#L168-L169

src/plugins/minimal_scene/MinimalScene.cc
ahcorde: There is only one mouseEvent, it's used for press, hover and release. This is required to know where was the mouse released.

chapulina: But we could also populate it on `RenderWindowItem::mousePressEvent`, right?

ahcorde: It's done 

https://github.com/ignitionrobotics/ign-gui/blob/bde1977b4600602fb63f08d74cfbeeaf5bd3919b/src/plugins/minimal_scene/MinimalScene.cc#L812-L815

src/plugins/minimal_scene/MinimalScene.cc
chapulina: Ah thanks. But it's strange that these are not done in the same place. The right click is in `Broadcast` while the left click is in `HandleMouseViewControl`. Do you know why they're not handled the same way?

src/plugins/minimal_scene/MinimalScene.cc
chapulina: Ah ok, so what I don't understand is why we'd set `PressPos` during the mouse release. AFAICT, we store the press position so that we can later compare it to the release position.

What's probably happening is that we're losing the original press position when we convert, right? Maybe we should store the press pos before convert and then set that. i.e.

```cpp
auto pressPos = this->dataPtr->mouseEvent.PressPos();
this->dataPtr->mouseEvent = convert(*_e);
this->dataPtr->mouseEvent.SetPressPos(pressPos);
```

chapulina (approved):
The context menu is working on `ign-gazebo` with this, let's get it in!

PR #286, opened 2021-09-17
Fix render engine GUI name argument to be setted in the minimal scene

ahcorde:

# 🦟 Bug fix

## Summary

This PR will set the render engine GUI name as a property of the MainWindow. Then this could be consume by the plugins. Right now the minimal_scene only works with "ogre2" (as a default render engine) this PR fix this and allow to set the render engine not only with the sdf but with the command line argument `--render-engine-gui`

## Checklist
- [ ] Added tests


include/ignition/gui/Helpers.hh
```diff
@@ -79,6 +79,15 @@ namespace ignition
     IGNITION_GUI_VISIBLE
     QStringList worldNames();
 
+    /// \brief The main window's "renderEngineGUI" property may be filled with
+    /// a string of the render engine gui name. This information can be
+    /// used by plugins to set the GUI render engine
+    /// This helper function provides a handy access to the render engine GUI
+    /// name
+    /// \return rneder engine GUI name, as stored in the `MainWindow`'s
```
chapulina, line 87:
```suggestion
    /// \return Name of render engine used on the GUI, as stored in the `MainWindow`'s
    /// "renderEngineGui" property.
```

src/MainWindow.cc
```diff
@@ -859,3 +859,9 @@ void MainWindow::SetShowDialogOnExit(bool _showDialogOnExit)
   this->dataPtr->showDialogOnExit = _showDialogOnExit;
   this->ShowDialogOnExitChanged();
 }
+
+/////////////////////////////////////////////////
+void MainWindow::SetRenderEngineGUI(std::string _renderEngineGUI)
```
chapulina, line 864:
```suggestion
void MainWindow::SetRenderEngineGUI(const std::string &_renderEngineGUI)
```

src/MainWindow.cc
```diff
@@ -859,3 +859,9 @@ void MainWindow::SetShowDialogOnExit(bool _showDialogOnExit)
...
+{
+  this->setProperty("renderEngineGUI", _renderEngineGUI.c_str());
```
chapulina, line 866: I think that `GUI` is a bit redundant here. This could be just `renderEngine`

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -908,18 +909,27 @@ void MinimalScene::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
...
+    {
+      if (nullptr != elem && nullptr != elem->GetText())
+      {
+        renderWindow->SetEngineName(elem->GetText());
+        // there is a problem with displaying ogre2 render textures that are in
+        // sRGB format. Workaround for now is to apply gamma correction
+        // manually.
+        // There maybe a better way to solve the problem by making OpenGL calls.
+        if (elem->GetText() == std::string("ogre2"))
+          this->PluginItem()->setProperty("gammaCorrect", true);
```
chapulina, line 927: Should this be moved out of this `if`? We need to set `gammaCorrect` for `ogre2` regardless of whether the engine name came from the window property or the XML.

chapulina (changes requested):
The approach looks good to me. Can we add tests for the new functions?

chapulina (approved):
I pushed some tweaks in d057f65ce1e83a34e20e8577d0554fb45f6e607f, LGTM. Feel free to merge if you and CI are happy with the changes :)

PR #287, opened 2021-09-19
Fix memory leak in ImageDisplay.cc

darksylinc:
## Summary

**Note:** This is a big memory leak in the order of 30 MB/s or more depending on many conditions; thus it should be prioritized. A 32GB RAM machine can run out of memory after approximately 20 min of real time simulation.

I don't like the interface of common::Image outputting a raw pointer that is deleted if the input is non-nullptr but it is up to the caller to the delete if it's not called again. And it's not documented either.
I don't like the amount of unnecessary copies required by `ImageDisplay::ProcessImage` either. Or the fact that we do one allocation per frame (instead of trying to keep an allocated region while in use).

*But all those are problems for another day*.

Don't forget to merge this to ign-gui5 and later :)

# 🦟 Bug fix

Fixes ignitionrobotics/ign-gazebo#1011

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -212,6 +212,8 @@ void ImageDisplay::ProcessImage()
         image.setPixel(i, j, value);
       }
     }
+
+    delete [] data;
   }
 
   this->dataPtr->provider->SetImage(image);


ahcorde (approved):

chapulina (approved):
Thanks for the fix!

osrf-triage:

PR #306, opened 2021-10-14
Support emitting an event on play/pause/step

adlarkin:
Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

# 🎉 New feature

Part of https://github.com/ignitionrobotics/ign-gazebo/issues/1106

Requires https://github.com/ignitionrobotics/ign-msgs/pull/190

## Summary
If the GUI's ECM is modified directly while simulation is paused, the GUI needs to tell the server how the ECM changed. The approach taken is to have the `WorldControl` service called in `ign-gazebo`'s `GuiRunner` instead of `ign-gui`'s `WorldControl` plugin. When play/pause/step is pressed in the GUI, an event is emitted that contains the `WorldControl` information. This event is processed by `ign-gazebo`'s `GuiRunner`. The `GuiRunner` will then take the information from this event, combine it with the GUI's ECM state, and call the `WorldControl` service that was being called in `ign-gui` before this PR. `ign-gui`'s `WorldControl` will listen to the world's `stats` topic for confirmation that the emitted `WorldControl` event was processed by the server.

It's also worth noting that the client/server communication that exists before this PR (direct service call) is still supported - see the `Test It` section below.

## Test it
_note: this PR needs to be tested with https://github.com/ignitionrobotics/ign-gazebo/pull/1109_

Run a gazebo world and try to play/pause/step from the GUI. You should see proper update behavior. You can also try playing/pausing from the command line to verify that those updates work properly as well.

To play from the command line with the original service (make sure you update `/world/shapes/control` to the service name that matches your world):
```
ign service -s /world/shapes/control  --reqtype  ignition.msgs.WorldControl --reptype ignition.msgs.Boolean --timeout 2000 --req 'pause: false'
```

To pause from the command line with the original service (make sure you update `/world/shapes/control` to the service name that matches your world):
```
ign service -s /world/shapes/control  --reqtype  ignition.msgs.WorldControl --reptype ignition.msgs.Boolean --timeout 2000 --req 'pause: true'
```

To play from the command line with the new service that's called by the emitted event (make sure you update `/world/shapes/control/state` to the service name that matches your world):
```
ign service -s /world/shapes/control/state  --reqtype  ignition.msgs.WorldControlState --reptype ignition.msgs.Boolean --timeout 2000 --req 'world_control: {pause: false}'
```

To pause from the command line with the new service that's called by the emitted event (make sure you update `/world/shapes/control/state` to the service name that matches your world):
```
ign service -s /world/shapes/control/state  --reqtype  ignition.msgs.WorldControlState --reptype ignition.msgs.Boolean --timeout 2000 --req 'world_control: {pause: true}'
```

Testing server/client communication through either the new event approach or traditional service approach can be toggled by adding the following to the `WorldControl` plugin's XML (`true` for event, `false` for service - see an example [here](https://github.com/ignitionrobotics/ign-gazebo/blob/e61eef90a392896b85253ee971098f1eafe56162/examples/worlds/camera_sensor.sdf#L108)):
```
<use_event>true</use_event>
```

## Checklist
- [x] Added tests


src/plugins/world_control/WorldControl.cc
```diff
@@ -212,6 +219,26 @@ void WorldControl::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
     ignerr << "Failed to create valid topic for world [" << worldName << "]"
            << std::endl;
   }
+
+  // advertise a service that allows other processes to notify WorldControl
+  // that a change to the GUI's ECM took place
+  const std::string guiEcmChangesService = "guiEcmChanges";
```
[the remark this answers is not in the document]
adlarkin: no longer needed as of https://github.com/ignitionrobotics/ign-gui/pull/306/commits/915a1a6fda3cbbd2fdb0a4af3238520e7bc53a7c

adlarkin:
@nkoenig I have updated this PR to reflect the new design that was discussed in https://github.com/ignitionrobotics/ign-gazebo/pull/1109#issuecomment-953472995.

include/ignition/gui/GuiEvents.hh
```diff
@@ -449,6 +450,34 @@ namespace ignition
...
+        static const QEvent::Type kType = QEvent::Type(QEvent::MaxUser - 19);
+
+        /// \brief Get the WorldControl information
+        /// \return The WorldControl information
+        public: const msgs::WorldControl &WorldControlInfo() const;
+
+        /// \brief Get whether the event is associated with
+        /// a user selecting "play" or not
+        /// \return True if the event is associated with a user selection play.
+        /// False otherwise
```
chapulina, line 474: I think this could be a bit clearer. The first time I read it, I thought it would return the same as `WorldControlInfo().pause()` (i.e. false may mean pause or no action). But looking at the code, false here does mean pause. So I think it's worth documenting it:

```suggestion
        /// False if the user selected "pause".
```

src/plugins/world_control/WorldControl.cc
```diff
@@ -219,11 +227,21 @@ void WorldControl::ProcessMsg()
...
 
-  if (!this->dataPtr->pause && this->dataPtr->msg.paused())
+  // If the pause state of the message doesn't match the pause state of this
+  // plugin, then play/pause must have occurred elsewhere (for example, the
+  // command line). If the pause state of the message matches the pause state
+  // of this plugin, but the pause state of the message differs from the
+  // previous message's pause state, this means that a pause/play request from
+  // this plugin has been registered by the server
+  if ((!this->dataPtr->pause && this->dataPtr->msg.paused()) ||
+      (!this->dataPtr->lastStatsMsgPaused && this->dataPtr->msg.paused()))
```
chapulina, line 237: Both these conditions have `&& this->dataPtr->msg.paused()` - can that be moved out to make it a bit simpler? For example

```
  if (this->dataPtr->msg.paused() &&
      (!this->dataPtr->pause || !this->dataPtr->lastStatsMsgPaused))
```

The same goes for the condition below. I'm having a hard time parsing it even with the comments.

src/plugins/world_control/WorldControl.cc
```diff
@@ -238,33 +256,21 @@ void WorldControl::OnWorldStatsMsg(const ignition::msgs::WorldStatistics &_msg)
...
-    if (_result)
-      QMetaObject::invokeMethod(this, "playing");
-  };
-
-  ignition::msgs::WorldControl req;
-  req.set_pause(false);
+  ignition::msgs::WorldControl msg;
+  msg.set_pause(false);
   this->dataPtr->pause = false;
-  this->dataPtr->node.Request(this->dataPtr->controlService, req, cb);
```
chapulina, line 251: `ign-gazebo` is not the only application using these plugins, so we should be mindful of other downstream users when changing behaviours. Can we do both (service + event), and offer the user an option to disable one or the other via XML?

chapulina (changes requested):

include/ignition/gui/GuiEvents.hh
adlarkin: False can mean more than just pause - for example, in this PR, false is also used for step: https://github.com/ignitionrobotics/ign-gui/blob/915a1a6fda3cbbd2fdb0a4af3238520e7bc53a7c/src/plugins/world_control/WorldControl.cc#L288

So, what if I said something like: `False if the user did not select play (some examples might be the user selecting "pause" or "step")`.

I don't know if we should document that false can only mean "pause", because perhaps downstream users have other interpretations of what false could be.

chapulina: > for example, in this PR, false is also used for step: 

Humm but the meaning is still that simulation should not continue running after receiving this command, right? I don't think that's different from saying that means pause :thinking:

src/plugins/world_control/WorldControl.cc
adlarkin: Yeah, thanks for the reminder - I have added support for event or service in https://github.com/ignitionrobotics/ign-gui/pull/306/commits/7ff16ce47ffbb1e0dc37d005333a4111a13e505d. The following can be added to `ign-gazebo`'s `gui.config` if you want to use a service (event is now the default):

```diff
diff --git a/src/gui/gui.config b/src/gui/gui.config
index 36166061..28bfedf9 100644
--- a/src/gui/gui.config
+++ b/src/gui/gui.config
@@ -137,6 +137,7 @@
   <step>true</step>
   <start_paused>true</start_paused>
 
+  <use_event>false</use_event>
 </plugin>
 
 <!-- World statistics -->
```

Does the migration guide need to be updated? If so, is that an update that should take place here, or in `ign-gazebo`?

src/plugins/world_control/WorldControl.cc
adlarkin: Thanks for the suggestion! I have applied it in https://github.com/ignitionrobotics/ign-gui/pull/306/commits/7ff16ce47ffbb1e0dc37d005333a4111a13e505d. Hopefully it's clearer now :crossed_fingers: :grimacing:

include/ignition/gui/GuiEvents.hh
adlarkin: So, after addressing https://github.com/ignitionrobotics/ign-gazebo/pull/1109#discussion_r740421855, this event actually does not need the `Play` method anymore. `ign-gazebo` has everything it needs from just the `msgs::WorldControl` piece. So, I've removed the `Play` method in https://github.com/ignitionrobotics/ign-gui/pull/306/commits/7ff16ce47ffbb1e0dc37d005333a4111a13e505d. I'll mark this as resolved, but feel free to re-open it if you have any more comments/questions.

adlarkin:
@chapulina the tests for the `WorldControl` plugin use the service approach instead of the new event approach (see https://github.com/ignitionrobotics/ign-gui/pull/306/commits/de55398da0c18286d8c19241586ae590a12daf4d). Do tests need to be added for the event approach as well? If so, I'm not sure how this would be tested since the receiving end of the service is in https://github.com/ignitionrobotics/ign-gazebo/pull/1109.

src/plugins/world_control/WorldControl.cc
nkoenig: I changed the default to use the service so that we don't break downstream users with a behavior change. On the Gazebo side, I set the default gui configs to set `<use_event>true</use_event>` in https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/ae91a9dab88d717124de89a1fd6d1ee5a2045e3b

adlarkin: > I changed the default to use the service so that we don't break downstream users with a behavior change. On the Gazebo side, I set the default gui configs to set `<use_event>true</use_event>`

That sounds good, thanks for making the change. Since behavior is still preserved for downstream users, I think we can mark this as resolved.

nkoenig:
I think the `step` button on the GUI now toggles the `Play` button.

adlarkin:
> I think the `step` button on the GUI now toggles the `Play` button.

Are you referring to what I'm seeing in https://github.com/ignitionrobotics/ign-gazebo/pull/1109#discussion_r740683720?

adlarkin:
> > I think the `step` button on the GUI now toggles the `Play` button.
> 
> Are you referring to what I'm seeing in [ignitionrobotics/ign-gazebo#1109 (comment)](https://github.com/ignitionrobotics/ign-gazebo/pull/1109#discussion_r740683720)?

This should be fixed now - see https://github.com/ignitionrobotics/ign-gazebo/pull/1109#discussion_r741489227

chapulina (commented):
> Do tests need to be added for the event approach as well? 

That would be ideal

> If so, I'm not sure how this would be tested since the receiving end of the service is in ignitionrobotics/ign-gazebo#1109.

How about listening to the event on the test and check that it is called? Maybe you can create a simple class that installs a filter :thinking:

src/plugins/world_control/WorldControl.cc
```diff
@@ -212,18 +228,44 @@ void WorldControl::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
     ignerr << "Failed to create valid topic for world [" << worldName << "]"
            << std::endl;
   }
+
+  // TODO(anyone) remove the option to use a service in ign-gui7 and use an
```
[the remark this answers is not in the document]
adlarkin: No, it's not necessary. I assumed we wouldn't want to maintain both long-term, but I'm okay with keeping the flexibility. I removed deprecation notes in https://github.com/ignitionrobotics/ign-gui/pull/306/commits/0112b87b05b1c11db48afcd0b977c87df9328188 and https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/e61eef90a392896b85253ee971098f1eafe56162.

adlarkin:
> How about listening to the event on the test and check that it is called? Maybe you can create a simple class that installs a filter thinking

I've added a test for the event approach in https://github.com/ignitionrobotics/ign-gui/pull/306/commits/84887d00cbfb5ce59a751f484f76550c564caf11.

nkoenig (approved):
Works for me. I'd like to switch out the header stepping flag in Garden. Here is the start: https://github.com/ignitionrobotics/ign-msgs/pull/199.

nkoenig:
@adlarkin , @chapulina would it be possible to get this in today so that I can make a release and unblock https://github.com/ignitionrobotics/ign-gazebo/pull/1157?

adlarkin:
> would it be possible to get this in today

That's okay with me - once @chapulina takes a look and no more changes are requested, feel free to merge.

I am seeing a new issue in Windows CI though that's related to the tests I added:
```
[703.453s] C:\Jenkins\workspace\ign_gui-pr-win\ws\build\ignition-gui6\src\plugins\world_control\moc_WorldControlEventListener.cpp(66,99): error C2491: 'ignition::gui::WorldControlEventListener::staticMetaObject': definition of dllimport static data member not allowed [C:\Jenkins\workspace\ign_gui-pr-win\ws\build\ignition-gui6\src\plugins\world_control\WorldControl.vcxproj]
```

Does anyone know how to fix this?

nkoenig:
> ```
> [703.453s] C:\Jenkins\workspace\ign_gui-pr-win\ws\build\ignition-gui6\src\plugins\world_control\moc_WorldControlEventListener.cpp(66,99): error C2491: 'ignition::gui::WorldControlEventListener::staticMetaObject': definition of dllimport static data member not allowed [C:\Jenkins\workspace\ign_gui-pr-win\ws\build\ignition-gui6\src\plugins\world_control\WorldControl.vcxproj]
> ```
> 
> Does anyone know how to fix this?

All of your tests were set to run only on linux, so I disable their compilation on windows in 80a64f4.

nkoenig:
Alright, CI is green-enough. There is a flaky test that has failed on Focal, but has passed in other iterations.

nkoenig:
@chapulina , I'm going to merge this by 10am today unless you stop me.

include/ignition/gui/GuiEvents.hh
```diff
@@ -27,6 +27,7 @@
 #include <ignition/common/MouseEvent.hh>
 #include <ignition/math/Vector2.hh>
 #include <ignition/math/Vector3.hh>
+#include <ignition/msgs.hh>
```
chapulina, line 30: It's preferable to include just the message that's needed, this helps with compile times.

src/plugins/world_control/WorldControl.cc
```diff
@@ -238,33 +274,47 @@ void WorldControl::OnWorldStatsMsg(const ignition::msgs::WorldStatistics &_msg)
...
-      QMetaObject::invokeMethod(this, "playing");
-  };
-
-  ignition::msgs::WorldControl req;
-  req.set_pause(false);
+  ignition::msgs::WorldControl msg;
+  msg.set_pause(false);
   this->dataPtr->pause = false;
-  this->dataPtr->node.Request(this->dataPtr->controlService, req, cb);
+  if (this->dataPtr->useEvent)
```
chapulina, line 280: This block from 280 to 294 is repeated 3 times in different functions, consider moving it to a separate function to reduce duplication

src/plugins/world_control/WorldControlEventListener.hh
```diff
@@ -0,0 +1,58 @@
...
+#include "ignition/gui/qt.h"
+
+namespace ignition
+{
+namespace gui
+{
+  /// \brief Helper class for testing listening to events emitted by the
+  /// WorldControl plugin. This is used for testing the event behavior of
+  /// the WorldControl plugin.
+  class IGNITION_GUI_VISIBLE WorldControlEventListener : public QObject
```
chapulina, line 33: Great work putting the tests together! Ideally we wouldn't include this as part of the plugin itself, since it's only used for testing. But I think this can be addressed later.

src/plugins/world_control/CMakeLists.txt
```diff
@@ -1,9 +1,19 @@
-ign_gui_add_plugin(WorldControl
-  SOURCES
-    WorldControl.cc
-  QT_HEADERS
-    WorldControl.hh
-  TEST_SOURCES
-    WorldControl_TEST.cc
-)
-
+if (NOT MSVC)
```
chapulina, line 1: Just curious, why did you need to separate Windows here? We're already disabling the test on Windows in the test's body.

chapulina (approved):
I just have minor comments, no blockers

src/plugins/world_control/CMakeLists.txt
nkoenig: Because Windows would still compile the testing framework and produce compile time errors.

chapulina: I see. It was compiling before. Not sure what causes the compiler error here, but I think it's ok to address this later.

src/plugins/world_control/WorldControlEventListener.hh
nkoenig: Done in 8e7cb8f

src/plugins/world_control/WorldControl.cc
nkoenig: Done in 8e7cb8f

include/ignition/gui/GuiEvents.hh
nkoenig: Done in 8e7cb8f

src/plugins/world_control/CMakeLists.txt
adlarkin: As a follow-up, yes,the tests themselves ignore windows. For example: https://github.com/ignitionrobotics/ign-gui/blob/ignition-gui6_6.0.0/src/plugins/world_control/WorldControl_TEST.cc#L66

However, the new helper test class that was added for checking events (https://github.com/ignitionrobotics/ign-gui/pull/306#issuecomment-959731077) has to be added as a QT header, or else there will be linking errors. So, it seemed like this new QT header is what was causing issues on windows, and @nkoenig made a patch in https://github.com/ignitionrobotics/ign-gui/pull/306#issuecomment-961378270 to prevent compiling the helper test class header on windows. I'm not sure if there's a better way to handle this or not, but hopefully that provides some more context/insights about this change.

nkoenig: I tried a few different options to make the header compile on windows, and gave up.

PR #307, opened 2021-10-30
[WIP] Performance improvements for marker.

arjo129:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
This PR is a WIP, however while working with markers I found that the experience was sub-optimal. This is largely due to performance issues in the `MarkerManager.cc `. Looking at the code, I've noticed several issues which can be addressed. The initial commit changes the data structure used for storing markers from a `map` which has `O(logn)` insertion/lookup to an `O(1)` `sts::unordered_map`. However there are many more improvements I can think of. I'm just outlining these here to ensure that Iwe revisit it at some point. These include:
- Change the lifetime to use a `priority_queue` to track items which are due to expire iunstead of the current O(n) loop.
- Change the `std::list` to an `std::deque` for better cache locality (will need profile + benchmark to confirm this actually works.)
- Change the id allocation algorithm to be more deterministic. Possibly by keeping track of a max id thus making the allocation more robust.
- Use a virtual marker list for CRUD operations (thus reducing calls to the underlying render processes) and moving the load of the main UI thread.
- Rate limitting the markers so if there is an excess of markers published we don't end up blocking the main UI thread.

## TODO

- [ ] write benchmarks.

## Checklist
- [ ] Added tests


arjo129:
Will open a new request which supercedes this pr.

PR #319, opened 2021-11-24
use qmldir to define QML module with IgnSpinBox

### LINKED ISSUE #310: Provide reusable QML components with common widgets
Widgets created with Ignition GUI often have duplicate code for interfaces to vectors, poses, colors, etc. It would be easier to construct various plugins if we offered these building blocks out of the box. It would also help users to write widgets that have the same look and feel.

## Desired behavior

When I'm writing a plugin, I'd like to import Ignition GUI's resources and immediately include common components. For example:

```
...
import "qrc:/qml"
...
IgnPose3d {
  id: myPose
  xValue: posX
  ...
}
```

## Alternatives considered

Keep using custom components for each plugin.

## Implementation suggestion

See the implementation for [IgnSpinBox](https://github.com/ignitionrobotics/ign-gui/blob/ign-gui6/include/ignition/gui/qml/IgnSpinBox.qml), which is currently used in multiple plugins.

## Additional context

Igntiion Gazebo's [ComponentInspector](https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo6/src/gui/plugins/component_inspector) plugin already implements a lot of these. The work should consist in porting them to `ign-gui` and making them as general as possible.

zflat:
# 🎉 New feature

Closes #310 

## Summary

This change adds a qmldir file that defines an importable QML module for common QML components implemented in `include/ignition/gui/qml/`.

With this change, it will be possible to import QML components using module import statements instead of file path.

Updating import statements will look like this:

```diff
-import "qrc:/qml"
+import ignition.gui 1.0
```

## Test it

* Build and run the application `ign run`
* Add the "Teleop" plugin
* Verify that the plugin loads and that the `IgnSpinBox` is part of the UI

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gui/Application.hh
+++ b/include/ignition/gui/Application.hh
@@ -46,6 +46,10 @@ namespace ignition
     class MainWindow;
     class Plugin;
 
+
+    /// /brief Import path for ign-gui QML modules added to the Qt resource system
+    const char *const QML_QRC_IMPORT_PATH = "qrc:/ign-gui-qml/";
+
     /// \brief Type of window which the application will display
     enum class WindowType : int
     {
--- a/include/ignition/gui/Application.hh
+++ b/include/ignition/gui/Application.hh
@@ -46,10 +46,6 @@ namespace ignition
     class MainWindow;
     class Plugin;
 
-
-    /// /brief Import path for ign-gui QML modules added to the Qt resource system
-    const char *const QML_QRC_IMPORT_PATH = "qrc:/ign-gui-qml/";
-
     /// \brief Type of window which the application will display
     enum class WindowType : int
     {
--- a/include/ignition/gui/Helpers.hh
+++ b/include/ignition/gui/Helpers.hh
@@ -79,6 +79,16 @@ namespace ignition
     IGNITION_GUI_VISIBLE
     QStringList worldNames();
 
+
+    /// /brief Import path for ign-gui QML modules added to the Qt resource system
+    /// This helper function returns the QRC resource path where custom ignition QML
+    /// modules can be imported from. To import an ignition QML module, add this path
+    /// to the QML engine's import path list before attempting to load a QML file
+    /// that imports ignition QML modules.
+    /// \return Resousrce path prefix as a string
+    IGNITION_GUI_VISIBLE
+    const QString qmlQrcImportPath();
+
     /// \brief Returns the first element on a QList which matches the given
     /// property.
     /// \param[in] _list The list to search through.
--- a/include/ignition/gui/Helpers.hh
+++ b/include/ignition/gui/Helpers.hh
@@ -80,7 +80,7 @@ namespace ignition
     QStringList worldNames();
 
 
-    /// /brief Import path for ign-gui QML modules added to the Qt resource system
+    /// \brief Import path for ign-gui QML modules added to the Qt resource system
     /// This helper function returns the QRC resource path where custom ignition QML
     /// modules can be imported from. To import an ignition QML module, add this path
     /// to the QML engine's import path list before attempting to load a QML file
--- a/include/ignition/gui/resources.qrc
+++ b/include/ignition/gui/resources.qrc
@@ -20,4 +20,8 @@
   <file>qml/images/menu.png</file>
   <file>qml/images/search.svg</file>
 </qresource>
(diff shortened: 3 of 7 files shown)


include/ignition/gui/Application.hh
```diff
@@ -46,6 +46,10 @@ namespace ignition
     class MainWindow;
     class Plugin;
 
+
+    /// /brief Import path for ign-gui QML modules added to the Qt resource system
+    const char *const QML_QRC_IMPORT_PATH = "qrc:/ign-gui-qml/";
```
chapulina, line 51: Can this be a member variable to `ApplicationPrivate`, or do we expect downstream applications to need it?

include/ignition/gui/resources.qrc
```diff
@@ -23,4 +23,8 @@
   <file>qml/images/export_icon.png</file>
   <file>qml/images/search.svg</file>
 </qresource>
+<qresource prefix="ign-gui-qml/ignition/gui">
```
chapulina, line 26: Can we add a comment here that we should add files here if they're meant to be used by downstream?

chapulina (commented):
Nice, thank you for cleaning it up!

The approach looks good to me. I also checked that `import "qrc:/qml"` still works, so this doesn't break any existing users.

Would you be able to retarget this PR to `ign-gui3`? Then we can merge it forward to all supported versions from there.

Thanks!

zflat:
> Nice, thank you for cleaning it up!
> 
> The approach looks good to me. I also checked that `import "qrc:/qml"` still works, so this doesn't break any existing users.
> 
> Would you be able to retarget this PR to `ign-gui3`? Then we can merge it forward to all supported versions from there.
> 
> Thanks!

I branched off of `ign-gui3` and force-pushed then updated this PR to target `ign-gui3`.

include/ignition/gui/resources.qrc
zflat: Added a comment in 26698357a3976e165f8a515afd3ce60fde1f6fbe but can add more explanation if needed. Let me know if you have any suggestions on additional information to include in the comment.

include/ignition/gui/Application.hh
zflat: I was thinking about downstream applications when I made this a constant.

chapulina: Got it. I think it's best to avoid the global variable though. How about one of these:

1. making it a private variable and adding an accessor function (i.e. `Application::QmlQrcImportPath`
2. creating a global function such as `ignition::gui::qmlQrcImportPath()`, maybe in `Helpers.hh`?

I think it would also be good to use a more modern type to hold this value. How about an `std::string`? I can also see the case for `QString` since that's what `addImportPath` takes.

zflat: I changed it to a global helper function in 8f4601832e07a1d95bff84bda3c0ba57b7a0c7d7

zflat:
@chapulina Are there any more QML components that should be added to the `ignition.gui 1.0` QML module in this pull request? Anything that has an `Ign` prefix? 
* IgnCard
* IgnCardSettings
* ...
* IgnSplit

Or should additional components be added in a separate pull request?

chapulina:
> Are there any more QML components that should be added to the ignition.gui 1.0 QML module in this pull request?

I was thinking about that. The only one that we really use downstream is `IgnSpinBox`, so I think that's the only one we really need to add right now. If users need to use the others, I think we can add them later.

There's been conversations about adding generic widgets for pose, vector, etc, which we repeat a lot in `ign-gui` and `ign-gazebo`. Once those are added I'd expect to put them in this module too.

include/ignition/gui/Helpers.hh
```diff
@@ -79,6 +79,16 @@ namespace ignition
     IGNITION_GUI_VISIBLE
     QStringList worldNames();
 
+
+    /// /brief Import path for ign-gui QML modules added to the Qt resource system
```
chapulina, line 83:
```suggestion
    /// \brief Import path for ign-gui QML modules added to the Qt resource system
```

chapulina (approved):
LGTM, are you planning on making more changes?

jennuine (approved):

osrf-triage:

osrf-triage:

osrf-triage:

Issue #54, opened 2020-04-25
Allow placing plugins on top of main toolbar


chapulina:
Currently, floating plugins can only be placed below the toolbar.

It would be nice to also support adding small plugins, usually consisting of some buttons, on top of the toolbar. We may be able to do this with the current floating plugins, just making sure that's a valid XY position and that their Z can be above the toolbar.

PR #335, opened 2021-12-14
Added a button that allows shutting down both the client and server.

peci1:
# 🎉 New feature

## Summary
Adds a button that allows closing the server by calling `/server_control::stop` service (in case the server runs in a different manager or on a different machine than the gui).

Depends on https://github.com/ignitionrobotics/ign-gazebo/pull/1240 to be able to actually stop the server.

## Test it

    ign gazebo -v4 -s &
    ign gazebo -v4 -g
    # add the Shutdown plugin
    # click the cross
    # both the GUI and server process should be stopped

## Checklist
- [x] Added tests


peci1:
What's weird is that even if I run just the GUI and no server, the `node->Request()` returns true. How is that possible?

peci1:
Please, also check the QML file after me. I haven't worked with QT quick before, so I pretty much don't know what I'm doing and I just copied the file from WorldControl.

jennuine (commented):
LGTM, just some minor comments.

Also, instead of creating a new plugin, would it be better to extend the close dialog menu (https://github.com/ignitionrobotics/ign-gui/pull/225)? We could modify it to say something like, "Cancel", "Shutdown all", "Close GUI only". What are your thoughts @chapulina ?

peci1:
Thanks for the nits, @jennuine . Regarding the close dialog extension - in my use-case, I want to run the app full-screen and without a window manager, so there is no window close button, and I also hide the hamburger menu with the Quit option. So having a separate plugin providing the close button inside the app is a good way for me.

Maybe the button in the plugin could just trigger the close dialog? If so, it'd be nice to be able to configure the available options for the dialog - e.g. in my use case I would not want to offer to close the GUI only and leave the server running. Would it also be possible to pass more configuration to the dialog, like the name of the server control service?

chapulina:
> instead of creating a new plugin, would it be better to extend the close dialog menu 

Yeah I also feel like a plugin to shutdown the application is a bit uncommon.

> in my use-case, I want to run the app full-screen and without a window manager, so there is no window close button, and I also hide the hamburger menu with the Quit option. 

Sounds like a good use case for 

* https://github.com/ignitionrobotics/ign-gui/issues/54

This way, the plugin could be more integrated into the UI and look like a native button.

> Maybe the button in the plugin could just trigger the close dialog?

That sounds like a good way to reduce duplication, so we have one code path that leads to closing the application. 

> Would it also be possible to pass more configuration to the dialog, like the name of the server control service?

Those could come under `<window>`, the inconvenient thing is that this tag can only come from a config file at the moment, not from an SDF for example.

src/plugins/shutdown_button/ShutdownButton.cc
```diff
@@ -0,0 +1,124 @@
...
+*/
+
+#include "ShutdownButton.hh"
+
+#include <csignal>
+#include <string>
+
+#include <ignition/common/Console.hh>
+#include <ignition/common/StringUtils.hh>
+#include <ignition/common/Time.hh>
```
chapulina, line 25: Is `Time.hh` needed? That's deprecated on `ign-common4` and removed on `ign-common5` in favor of `chrono`.

peci1:
> > Would it also be possible to pass more configuration to the dialog, like the name of the server control service?
> 
> Those could come under `<window>`, the inconvenient thing is that this tag can only come from a config file at the moment, not from an SDF for example.

What? SDF can configure GUI?

peci1:
> Sounds like a good use case for #54
>
> This way, the plugin could be more integrated into the UI and look like a native button.

I'm not sure I can help with #54. So what if this PR makes the close button plugin and once #54 is resolved, it could just be adjusted to better fit the toolbar UI?

chapulina:
> What? SDF can configure GUI?

Just the plugins right now. The plugins inside `<world><gui>`, such as [these](https://github.com/ignitionrobotics/ign-gazebo/blob/0810a6b5aac153904e47b0ccbdd8e548a8a99d9f/examples/worlds/sensors_demo.sdf#L30-L200), are passed to `ign-gui` the same way as the `<plugin>` tags in a `.config` file.

> So what if this PR makes the close button plugin and once #54 is resolved, it could just be adjusted to better fit the toolbar UI?

Yeah I think that's reasonable.

src/plugins/CMakeLists.txt
```diff
@@ -120,6 +120,7 @@ add_subdirectory(key_publisher)
 add_subdirectory(publisher)
 add_subdirectory(scene3d)
 add_subdirectory(screenshot)
+add_subdirectory(shutdown_button)
```
chapulina, line 123: Alphabetize

examples/config/time.config
```diff
@@ -81,3 +81,14 @@
   <topic>/world_stats</topic>
 </plugin>
 
+<plugin filename="ShutdownButton">
```
chapulina, line 84: Can we create a new config file for it? It's better to keep the configs targeted

src/plugins/shutdown_button/ShutdownButton.cc
```diff
@@ -0,0 +1,124 @@
...
+
+/////////////////////////////////////////////////
+void ShutdownButton::OnStop()
+{
+  std::function<void(const ignition::msgs::Boolean &, const bool)> cb =
+      [](const ignition::msgs::Boolean &/*_rep*/, const bool /*_result*/)
+  {
+    ignmsg << "Request to shutdown the simulation server sent. Stopping client "
+           << "now." << std::endl;
+    std::raise(SIGTERM);
```
chapulina, line 104: Can we call `close()` on the window instead? `SIGTERM` feels a bit extreme if we can avoid it.

src/plugins/shutdown_button/ShutdownButton.hh
```diff
@@ -0,0 +1,81 @@
...
+#include <memory>
+
+#ifdef _MSC_VER
+#pragma warning(push, 0)
+#endif
+#include <ignition/msgs.hh>
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
+#include <ignition/transport.hh>
```
chapulina, line 30: These includes can be moved to the cc file. And ideally we'd include just what we need (i.e. the `pb.h` include and `Node.hh`)

src/plugins/shutdown_button/ShutdownButton.qml
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+import QtQuick 2.9
+import QtQuick.Controls 2.2
+import QtQuick.Controls.Material 2.1
+import QtQuick.Layouts 1.3
+import "qrc:/qml"
```
chapulina, line 21: I think this `import` isn't needed

chapulina (commented):
The general idea is ok to me, just have some minor comments

src/plugins/shutdown_button/ShutdownButton.qml
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2021 Open Robotics
```
jennuine, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

src/plugins/shutdown_button/ShutdownButton_TEST.cc
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (C) 2021 Open Robotics
```
jennuine, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

src/plugins/CMakeLists.txt
peci1: I don't understand... `screenshot < shutdown < teleop` ... or is it too late already?

examples/config/time.config
peci1: Done.

src/plugins/shutdown_button/ShutdownButton.cc
peci1: I no longer use signals for this.

src/plugins/shutdown_button/ShutdownButton.hh
peci1: ShutdownButton is now just a thin plugin calling the window close event. So most includes went away. Some functionality moved to MainWindow, and I made the includes as precise as possible.

src/plugins/shutdown_button/ShutdownButton.qml
peci1: yes, it isn't.

peci1:
I've pushed the new implementation which expands on the functionality of the exit confirmation dialog and makes ShutdownButton plugin just a thin layer that invokes this dialog (or performs its action if the dialog is not to be shown).

See `examples/config/dialog_on_exit_shutdown.config` for the newly added options.

Example run:

```
ign gazebo -v4 -s &
ign gazebo -v4 -g --gui-config examples/config/dialog_on_exit_shutdown.config
```

peci1:
Snímek obrazovky pořízený 2022-01-04 04-40-55

peci1:
Some things to consider improving:

- The alignment and order of the buttons in the dialog is not fixed - it is determined by the OS. That's pretty bad as e.g. the Linux order is weird. It is because I wanted to retain the `standardButtons` functionality of the dialog, so the "shutdown" button is represented as `DISCARD` type button (it actually makes sense, I think).
- I'm not sure if the duality of the possible content of `<dialog_on_exit>` is okay (it can be either boolean or a XML snippet). See the docs where I tried to describe it.
- Some strings that could possibly be localized in the future moved to .hh/.cc files from the QML (default texts of the buttons and prompt). As you can see from the screenshot, the Cancel button is localized by the OS, which creates a bit messy impression (it is no problem to also add configurable text of the cancel button, but it seemed unnecessary to me).
- I would like to automatically recognize that the GUI and server are in the same process and accordingly simplify the confirmation dialog (i.e. it's not possible to close only the GUI). But that would need to be done in ign-gazebo/src/ign.cc if I'm right?

jennuine (commented):
> I'm not sure if the duality of the possible content of <dialog_on_exit> is okay (it can be either boolean or a XML snippet). See the docs where I tried to describe it.

It is possible but IMO is getting unnecessarily complex. For the confirmation dialog, I think we should simplify things either by:

1. Always show all options (ie., cancel, quit server & gui, quit gui only) in the dialog when `<dialog_on_exit>true</dialog_on_exit>` is set
2. Add new element to the config, something like `<dialog_on_exit_extra>` or `<extend_dialog_on_exit>` which takes a `bool` (default would be false)

I'm leaning towards (2) because there may be users who do not know (or need to know) about the 2 processes and it would avoid confusion (especially when they are run in the same process; eg, "Shutdown GUI only was clicked but why did it shutdown the server?"). Do you have any thoughts @chapulina ?

examples/config/dialog_on_exit_shutdown.config
```diff
@@ -0,0 +1,27 @@
...
+  <width>1000</width>
+  <height>845</height>
+  <default_exit_action>SHUTDOWN_SERVER</default_exit_action>
+  <dialog_on_exit>
+    <show>true</show>
+    <prompt_text>really? žřčš</prompt_text>
+    <show_shutdown_button>true</show_shutdown_button>
+    <show_close_gui_button>true</show_close_gui_button>
+    <shutdown_button_text>Quit Server&amp;GUI</shutdown_button_text>
+    <close_gui_button_text>Quit GUI only</close_gui_button_text>
```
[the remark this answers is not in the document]
peci1: I am the user :) I'm building a kiosk running SubT simulator and the expectation is that it will be used by people who have no idea what runs under the hood. So showing them "Quit Server & GUI" would be really weird and I'd like to put there a simple "Exit" or something similar (I'm hiding the "Close GUI only" button as I always want to shutdown both; and I don't dare running both using `ign gazebo` as the GUI segfaults and would shot down the server cancelling the progress in the simulation; instead, I use good old `roslaunch` to keep the GUI alive =) ).

peci1:
> I'm leaning towards (2) because there may be users who do not know (or need to know) about the 2 processes and it would avoid confusion (especially when they are run in the same process; eg, "Shutdown GUI only was clicked but why did it shutdown the server?"). Do you have any thoughts @chapulina ?

(2) seems a good option. I'd just suggest a different name, what about `<dialog_on_exit_config>`?

(1) would not be an option for my use-case as I need to only allow the "shutdown gui & server" option (see the previous code comment).

Hiding the "Close GUI only" option when both are running in the same process or process manager is something that should really be done, but I did not want to clutter this PR with it. I'm not even sure which library would be the correct "entry point" for such implementation (probably somewhere around ign-gazebo/ign.cc?).

jennuine:
> (2) seems a good option. I'd just suggest a different name, what about <dialog_on_exit_config>?

How about `<shutdown_config>`?

> So showing them "Quit Server & GUI" would be really weird and I'd like to put there a simple "Exit" or something similar 

But instead of changing the button text through the config, we can check if the new element is set then update the text accordingly

> I'm not even sure which library would be the correct "entry point" for such implementation (probably somewhere around ign-gazebo/ign.cc?).

This I'm not sure about either but hopefully Louise will be able to point you in the right direction :)

peci1:
> How about `<shutdown_config>`?

That loses any connection to the dialog it configures, in my point of view...
 
> But instead of changing the button text through the config, we can check if the new element (in the config) is set then update the text accordingly through code

Sounds good. Is there an implementation I could get inspired from? As I'm pretty new to QtQuick/QML, I'm a bit helpless here...

jennuine:
> That loses any connection to the dialog it configures, in my point of view...

You're right. How about `<dialog_on_exit_options>`? I'm ok with `<dialog_on_exit_config>` if that is preferred.

> Sounds good. Is there an implementation I could get inspired from? 

Maybe this will help? https://github.com/ignitionrobotics/ign-gui/pull/225/files

Here's a quick suggestion summary (keep in mind it's not exact):
- Read the config for `<dialog_on_exit_x>`
- Store in boolean `Q_PROPERTY` which represent if the extra button / text will be shown
Eg.,
https://github.com/ignitionrobotics/ign-gui/blob/6db723c7781aca2af798d4da4cb3bec60c99b8d5/src/Application.cc#L278-L283
- In the qml you can change text based on the variable with lambdas
Eg.,
https://github.com/ignitionrobotics/ign-gui/blob/838271e7d87402ef7f2e82368b4a36d4c8c62b73/include/ignition/gui/qml/Main.qml#L347
 can become `title: (extraDialogOpts ? "Do you really want to exit?" : "How do you want to close?")`

--- 

Hope this helps :) let me know if you need me to clarify anything

peci1:
Thanks for the suggestions, jennuine. I went with `<dialog_on_exit_options>` and I like it much more than the previous dual approach. I also managed to move the default texts back to QML, without even needing to add more properties - I just check whether the string coming from configuration is non-empty.

chapulina (commented):
> I would like to automatically recognize that the GUI and server are in the same process and accordingly simplify the confirmation dialog (i.e. it's not possible to close only the GUI). But that would need to be done in ign-gazebo/src/ign.cc if I'm right?

Yeah that would need to be done on Gazebo's side, maybe after it creates the `MainWindow` on `Gui.cc`.

---

Thanks for iterating, @peci1, it's working well for me, I just have minor comments.

include/ignition/gui/MainWindow.hh
```diff
@@ -39,9 +39,18 @@ namespace ignition
...
   {
+    Q_NAMESPACE
     class MainWindowPrivate;
     struct WindowConfig;
 
+    /// \brief The action executed when GUI is closed without prompt.
+    enum class ExitAction
+    {
+      CLOSE_GUI,  //! \brief Close GUI and leave server running
+      SHUTDOWN_SERVER,  //! \brief Close GUI and shutdown server
```
chapulina, line 50: We've been doing this:

```suggestion
    /// \brief Close GUI and shutdown server
      SHUTDOWN_SERVER,  
```

examples/config/dialog_on_exit_shutdown.config
```diff
@@ -0,0 +1,27 @@
...
+<window>
+  <width>1000</width>
+  <height>845</height>
+  <default_exit_action>SHUTDOWN_SERVER</default_exit_action>
+  <dialog_on_exit>true</dialog_on_exit>
+  <dialog_on_exit_options>
+    <prompt_text>really? žřčš</prompt_text>
+    <show_shutdown_button>true</show_shutdown_button>
+    <show_close_gui_button>true</show_close_gui_button>
+    <shutdown_button_text>Quit Server&amp;GUI</shutdown_button_text>
```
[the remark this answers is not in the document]
peci1: That's weird, it works for me. I'll ditch the `&` and use `and` to avoid such problems.

include/ignition/gui/qml/Main.qml
```diff
@@ -325,24 +335,51 @@ ApplicationWindow
...
-    onAccepted: {
-      Qt.quit()
+    standardButtons: (exitDialogShowCloseGui ? Dialog.Ok : Dialog.NoButton) | (exitDialogShowShutdown ? Dialog.Discard : Dialog.NoButton) | Dialog.Cancel
+
+    // The button texts need to be changed later than in onCompleted as standardButtons change later
+    onAboutToShow: function () {
+      if (exitDialogShowCloseGui)
+        footer.standardButton(Dialog.Ok).text = (exitDialogCloseGuiText ? exitDialogCloseGuiText : "Close GUI")
+      if (exitDialogShowShutdown)
+        footer.standardButton(Dialog.Discard).text = (exitDialogShutdownText ? exitDialogShutdownText : "Shutdown simulation")
```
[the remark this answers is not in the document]
peci1: Hmm, but this is intended to shutdown both GUI and server... I wanted to keep the label short. But maybe "Shutdown GUI and server" would be okay.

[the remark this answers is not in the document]
peci1: Or "Close GUI and server"

peci1 (commented):
Thanks for the review, Louise. I'll work through it today.

src/plugins/shutdown_button/ShutdownButton.qml
```diff
@@ -0,0 +1,50 @@
...
+    text: closeIcon
+    checkable: true
+    Layout.alignment : Qt.AlignVCenter
+    Layout.minimumWidth: width
+    Layout.leftMargin: 10
+    onClicked: {
+      ShutdownButton.OnStop()
+    }
+    Material.background: Material.primary
+  }
```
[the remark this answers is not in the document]
peci1: I've put there "Quit" to be consistent with the hamburger menu entry.

tutorials/04_layout.md
```diff
@@ -28,7 +28,24 @@ by adding a `<window>` element to the config file. The child elements are:
                     the menu. If `from_paths` is true, all plugins will be shown
                     anyway, so adding `<show>` has no effect. For the plugin to
                     be shown, it must be on the path.
-* `<dialog_on_exit>`: If true, a confirmation dialog will show up when closing the window.
+* `<default_exit_action>`: Default `CLOSE_GUI`. If set to `SHUTDOWN_SERVER` and `<dialog_on_exit>`
+                           is `false`, closing the simulation window will stop the server even if it
```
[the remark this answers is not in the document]
peci1: Okay, I re-worded the docs.

include/ignition/gui/qml/Main.qml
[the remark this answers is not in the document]
peci1: I've put there "Shutdown server and GUI".

src/plugins/shutdown_button/ShutdownButton.cc
```diff
@@ -0,0 +1,53 @@
...
+/////////////////////////////////////////////////
+ShutdownButton::ShutdownButton() : Plugin()
+{
+}
+
+/////////////////////////////////////////////////
+ShutdownButton::~ShutdownButton() = default;
+
+/////////////////////////////////////////////////
+void ShutdownButton::LoadConfig(const tinyxml2::XMLElement */*_pluginElem*/)
```
chapulina, line 38: Windows is picky:

`'*/' found outside of comment`

```suggestion
void ShutdownButton::LoadConfig(const tinyxml2::XMLElement * /*_pluginElem*/)
```

peci1: Done.

peci1:
I added some tests. They're not exactly unit tests of the getters/setters as I find those useless (setters are setting... hmm). I instead wrote them as little integration tests spawning the GUI, displaying the dialog and looking for the elements that are expected to be found.

I had just one difficulty - no matter what I tried, I could not get the list line of `ExitDialogDefaultButtons` test succeed, so I commented it out with a TODO note. It examines whether the mainwindow stopped being visible after triggering the `clicked()` signal of one of the close buttons. However, the test tells the window is still there, even though I called `QCoreApplication::processEvents();` and waited a bit. So maybe I'm missing some other event processing step. But the signal gets called, because e.g. the server stopping request is published. Any thoughts?

include/ignition/gui/MainWindow.hh
```diff
@@ -360,6 +426,60 @@ namespace ignition
...
+        const QString& _exitDialogCloseGuiText);
+
+      /// \brief Get the topic of the server control service.
+      /// \return The service topic.
+      public: Q_INVOKABLE std::string ServerControlService() const;
+
+      /// \brief Set the topic of the server control service.
+      /// \param[in] _service The service topic.
+      public: Q_INVOKABLE void SetServerControlService(
+        const std::string& _service);
```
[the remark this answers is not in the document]
peci1: Added.

chapulina (commented):
> They're not exactly unit tests of the getters/setters as I find those useless (setters are setting... hmm)

You'd be surprised how many typo and copy/paste errors those catch :innocent: 

> instead wrote them as little integration tests spawning the GUI, displaying the dialog and looking for the elements that are expected to be found.

These are much nicer though, thanks!

>  the test tells the window is still there

I got the test to work with this, but it doesn't look pretty:

```diff
diff --git a/include/ignition/gui/qml/Main.qml b/include/ignition/gui/qml/Main.qml
index 7a12f382..70a33c87 100644
--- a/include/ignition/gui/qml/Main.qml
+++ b/include/ignition/gui/qml/Main.qml
@@ -54,6 +54,12 @@ ApplicationWindow
   property bool exitDialogShowCloseGui: MainWindow.exitDialogShowCloseGui
   property string exitDialogShutdownText: MainWindow.exitDialogShutdownText
   property string exitDialogCloseGuiText: MainWindow.exitDialogCloseGuiText
+
+  /**
+   * Flag to indicate if the close event was triggered by the close dialog.
+   */
+  property bool closingFromDialog: false
+
   /**
    * Tool bar background color
    */
@@ -83,7 +89,12 @@ ApplicationWindow
   onClosing: {
     close.accepted = !showDialogOnExit
     if(showDialogOnExit){
-      confirmationDialogOnExit.open()
+      if (closingFromDialog) {
+        close.accepted = true;
+      }
+      else {
+        confirmationDialogOnExit.open();
+      }
     } else if (defaultExitAction == ExitAction.SHUTDOWN_SERVER) {
       MainWindow.OnStopServer()
     }
@@ -373,7 +384,8 @@ ApplicationWindow
       DialogButtonBox {
         onClicked: function (btn) {
           if (btn == this.standardButton(Dialog.Ok)) {
-            Qt.quit()
+            closingFromDialog = true;
+            window.close();
           }
           else if (btn == this.standardButton(Dialog.Discard)) {
             MainWindow.OnStopServer()
@@ -381,7 +393,8 @@ ApplicationWindow
             timer.interval = 100;
             timer.repeat = false;
             timer.triggered.connect(function() {
-              Qt.quit()
+              closingFromDialog = true;
+              window.close();
             });
             timer.start();
           }
diff --git a/src/MainWindow_TEST.cc b/src/MainWindow_TEST.cc
index 357af4e6..6d6a4bd4 100644
--- a/src/MainWindow_TEST.cc
+++ b/src/MainWindow_TEST.cc
@@ -16,6 +16,7 @@
 */
 
 #include <gtest/gtest.h>
+#include <chrono>
 #include <thread>
 
 #include <QQmlProperty>
@@ -40,6 +41,7 @@ char* g_argv[] =
 
 using namespace ignition;
 using namespace gui;
+using namespace std::chrono_literals;
 
 /////////////////////////////////////////////////
 // See https://github.com/ignitionrobotics/ign-gui/issues/75
@@ -603,13 +605,18 @@ TEST(MainWindowTest,
   EXPECT_TRUE(mainWindow->QuickWindow()->isVisible());
   QMetaObject::invokeMethod(
     buttonRoles[ButtonRole::DestructiveRole], "clicked");
-  QCoreApplication::processEvents();
-  std::this_thread::sleep_for(std::chrono::milliseconds(150));
-  QCoreApplication::processEvents();
+
+  int sleep = 0;
+  const int maxSleep = 100;
+  for (; mainWindow->QuickWindow()->isVisible() && sleep < maxSleep; ++sleep)
+  {
+    std::this_thread::sleep_for(100ms);
+    QCoreApplication::processEvents();
+    sleep++;
+  }
+
   EXPECT_TRUE(shutdownCalled);
-  // TODO(peci1) Could not get this to work, probably some more event processing
-  //  is needed
-  // EXPECT_FALSE(mainWindow->QuickWindow()->isVisible());
+  EXPECT_FALSE(mainWindow->QuickWindow()->isVisible());
 }
 
 /////////////////////////////////////////////////
```

peci1:
> I got the test to work with this, but it doesn't look pretty:

I thinks it's a clean solution, I applied it.

Only the test calling server shutdown needed the wait, as the other closing gui only has no wait in the callback. I hope this works the same on all OSes.

I fixed the asterisk placement.

So I think this PR is ready to go :)

include/ignition/gui/qml/Main.qml
```diff
@@ -325,24 +344,59 @@ ApplicationWindow
...
+    standardButtons:
+      (exitDialogShowCloseGui ? Dialog.Ok : Dialog.NoButton) |
+      (exitDialogShowShutdown ? Dialog.Discard : Dialog.NoButton) |
+      Dialog.Cancel
+
+    // The button texts need to be changed later than in onCompleted as standardButtons change later
+    onAboutToShow: function () {
+      if (exitDialogShowCloseGui)
+        footer.standardButton(Dialog.Ok).text =
+          (exitDialogCloseGuiText ? exitDialogCloseGuiText : "Close GUI")
```
chapulina, line 375: Just a final nitpick, I was using `ign-gazebo` with this PR and noticed that it's a bit confusing to see this by default. Can we default to `Close`? If you're planning to follow up with an `ign-gazebo` PR this is less concerning, but I think that in general "GUI" can be a strange word to display to users by default for other applications too.

image

chapulina (approved):

jennuine (commented):
Looks good, left some minor comments

include/ignition/gui/qml/Main.qml
peci1: I understand your request to change the text, but imagine a dialog with "Cancel" and "Close". I hate these dialogs. Does "Close" close just the dialog, or the whole app? Before this PR, there was just "OK", which was good when there were just two options, but would be weird with three.

I could make the logic a bit more complicated and test whether there are just two, or all three buttons displayed. If only two were displayed, the text of the non-cancel button would be just "OK" (regardless if it is close gui or shutdown button). And if all three would be displayed, then the texts of the buttons could be as they are now. And in case any of the texts is customized, it would be used in all cases.

Another option would be to leave it completely on the users to redefine the button texts whenever they display all 3 buttons. But if only the shutdown button text were defined, it would lead to a dialog with "Shutdown server and GUI", "Cancel" and "OK" buttons, which is not best...

examples/config/dialog_on_exit_shutdown.config
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
+
+<!-- Window -->
+<window>
+  <width>1000</width>
+  <height>845</height>
+  <default_exit_action>SHUTDOWN_SERVER</default_exit_action>
+  <dialog_on_exit>true</dialog_on_exit>
+  <dialog_on_exit_options>
+    <prompt_text>really? žřčš</prompt_text>
```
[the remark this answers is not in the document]
peci1: I've put it there to make sure UTF-8 is interpreted correctly. But it can be removed now that we know it is... or it could stay so that everyone who runs the example will see that the config is UTF-safe.

examples/config/dialog_on_exit_shutdown.config
```diff
@@ -0,0 +1,27 @@
...
+    <show_shutdown_button>true</show_shutdown_button>
+    <show_close_gui_button>true</show_close_gui_button>
+    <shutdown_button_text>Quit Server and GUI</shutdown_button_text>
+    <close_gui_button_text>Quit GUI only</close_gui_button_text>
```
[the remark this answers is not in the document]
peci1: Well, the close_gui button is displayed by default, while the shutdown button is not. And so far empty `<*_text>` has meant to leave there the original text. So if we removed the `<show_*>` config options, I don't see a way to disable the close_gui button.

include/ignition/gui/qml/Main.qml
chapulina: > I hate these dialogs. 

Ah yeah good point, that's horrible.

> Another option would be to leave it completely on the users to redefine the button texts whenever they display all 3 buttons.

I think that's reasonable, if the user is customizing the dialog, they should be in a position to edit all buttons.

> "OK" (regardless if it is close gui or shutdown button)

Along the lines of the comment above, I think that's ok. When the user starts customizing the dialog, when may change the prompt and button texts as needed.

peci1: Good, I implemented your latest suggestion. I also added a few notes to docs and the tutorial, which should be enough to inform users of this API about the need of changing the default text value.

src/Application.cc
```diff
@@ -271,12 +273,89 @@ bool Application::LoadConfig(const std::string &_config)
     this->dataPtr->windowConfig.MergeFromXML(std::string(printer.CStr()));
 
     // Closing behavior.
+    if (auto defaultExitActionElem =
+      winElem->FirstChildElement("default_exit_action"))
+    {
+      ExitAction action{ExitAction::CLOSE_GUI};
+      const std::string value{defaultExitActionElem->GetText()};
+      if (value == "SHUTDOWN_SERVER")
```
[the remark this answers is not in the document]
peci1: done

examples/config/dialog_on_exit_shutdown.config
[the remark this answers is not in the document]
peci1: removed the diacritics characters

src/MainWindow.cc
```diff
@@ -161,6 +193,45 @@ void MainWindow::OnSaveConfigAs(const QString &_path)
...
 
+/////////////////////////////////////////////////
+void MainWindow::OnStopServer()
+{
+  std::function<void(const ignition::msgs::Boolean &, const bool)> cb =
+    [](const ignition::msgs::Boolean &_rep, const bool _result)
+    {
+      if (_rep.data() && _result)
+      {
+        ignmsg << "Simulation server acknowledged to be shut down."
```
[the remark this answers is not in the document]
peci1: changed

src/MainWindow.cc
```diff
@@ -161,6 +193,45 @@ void MainWindow::OnSaveConfigAs(const QString &_path)
...
+               << std::endl;
+      }
+      else
+      {
+        ignerr << "There was a problem instructing the simulation server to "
+               << "shut down. It may keep running." << std::endl;
```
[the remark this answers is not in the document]
peci1: changed

src/MainWindow.cc
```diff
@@ -859,3 +943,83 @@ void MainWindow::SetShowDialogOnExit(bool _showDialogOnExit)
...
+}
+
+/////////////////////////////////////////////////
+std::string MainWindow::ServerControlService() const
+{
+  return this->dataPtr->controlService;
+}
+
+/////////////////////////////////////////////////
+void MainWindow::SetServerControlService(const std::string& _service)
```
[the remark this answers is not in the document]
peci1: Found more of these and fixed.

src/MainWindow_TEST.cc
```diff
@@ -374,6 +379,335 @@ TEST(MainWindowTest,
...
+void FindExitDialogButtons(
+  MainWindow *_mainWindow,
+  std::unordered_set<ButtonRole> &_roles,
+  std::unordered_map<ButtonRole, QQuickItem*> &_buttonRoles)
+{
+  auto dialog = _mainWindow->QuickWindow()->findChild<QObject*>(
+    "confirmationDialogOnExit");
+  ASSERT_NE(nullptr, dialog);
+
+  QObject* buttonBox{nullptr};
```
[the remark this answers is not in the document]
peci1: fixed

jennuine:
Hmm, the `Quit Server and GUI` button isn't shutting down the server for me

peci1:
> Hmm, the `Quit Server and GUI` button isn't shutting down the server for me

Do you have up-to-date ign-gazebo 3? The shutdown functionality has only been implemented lately. It works for me locally and it also calls the correct service in tests, so it should work as long as the service server is implemented.

jennuine (approved):
> Do you have up-to-date ign-gazebo 3?
Whoops, I updated the wrong workspace. It works well for me, thanks for iterating!

peci1:
Hmm, I don't understand either of the CI failures...

chapulina:
> Hmm, I don't understand either of the CI failures...

I believe they're flaky due to
*  https://github.com/ignitionrobotics/ign-gui/issues/58

chapulina:
Thanks for iterating, merging!

osrf-triage:

osrf-triage:

PR #342, opened 2021-12-23
NavSat (GPS) map plugin

chapulina:
# 🎉 New feature

* Part of https://github.com/ignitionrobotics/ign-sensors/issues/23
* Requires https://github.com/ignitionrobotics/ign-msgs/pull/206
* Required by https://github.com/ignitionrobotics/ign-gazebo/pull/1248

## Summary

Ported @Sarath18 's GPS plugin from `ign-rviz` to `ign-gui`, see

* https://github.com/ignitionrobotics/ign-rviz/pull/66
* https://github.com/ignitionrobotics/ign-rviz/issues/65

I had to make a few changes to the QML for it to work for me on Bionic:

* I couldn't get the `mapbox` plugin to work, so I went with `osm`
* `supportedMapTypes` wasn't working for me so I removed it
* The unicode symbol `⌖` won't render for me when I load the GUI, mysterious... So I went with a different one for now which is not as nice: `✲`

I can think of a few different features to add, like showing multiple vehicles on the same map, marking the world origin, clicking on the map to choose the location to spawn a model... The possibilities are endless! This is just the first MVP.

## TODO

- [x] Fix warnings `[GUI] [Wrn] [Application.cc:674] [QT] file::/NavSatMap/NavSatMap.qml:97:3: QML Map: Binding loop detected for property "center"`

## Test it

It works with https://github.com/ignitionrobotics/ign-gazebo/pull/1248:

navsat_move

navsat_origin

## Checklist
- [ ] Added tests
- [x] Added example and/or tutorial - on Gazebo


src/plugins/navsat_map/NavSatMap.cc
```diff
@@ -0,0 +1,191 @@
...
+NavSatMap::~NavSatMap()
+{
+}
+
+/////////////////////////////////////////////////
+void NavSatMap::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
+{
+  // Default name in case user didn't define one
+  if (this->title.empty())
+    this->title = "Message display";
```
jennuine, line 76: Should the title be `"NavSatMap display"` instead?

src/plugins/navsat_map/NavSatMap.qml
```diff
@@ -0,0 +1,167 @@
...
+
+    onClicked: {
+      centering = true
+    }
+
+    hoverEnabled: true
+
+    ToolTip.delay: Qt.styleHints.mousePressAndHoldInterval
+    ToolTip.visible: hovered
+    ToolTip.text: qsTr("Keep point on center")
```
jennuine, line 157: When I first read this before clicking, I didn't understand what this meant. How about something like, `"Reset to center"` or `"Reset point to center"`?

jennuine (commented):
LGTM, left some minor comments

> TODO: Fix warnings [GUI] [Wrn] [Application.cc:674] [QT] file::/NavSatMap/NavSatMap.qml:97:3: QML Map: Binding loop detected for property "center"

I don't get this warning on focal :thinking: Although I do get it for:
```bash
[GUI] [Wrn] [Application.cc:674] [QT] file::/Gazebo/GazeboDrawer.qml:147:3: QML Dialog: Binding loop detected for property "implicitHeight"
```

src/plugins/navsat_map/NavSatMap.qml
chapulina: Yeah I agree it's confusing. Ignition RViz [just says "Center"](https://github.com/ignitionrobotics/ign-rviz/blob/632a18ade4ece9d0602a10ae48559c8b6d8392c4/ign_rviz_plugins/res/qml/GPSDisplay.qml#L213), so I added the "Keep point on" here in an attempt to make it clearer, but I agree it's still confusing.

I think "Reset" may give the wrong impression though, because it's not a one-time thing. Maybe something like "Track point", "Track location", "Enable tracking"...?

jennuine: Ah I see, I like "Track point" or "Track position"

src/plugins/navsat_map/NavSatMap.cc
chapulina: Ah nice catch! I changed it to "Navigation satellite map": c898d07

src/plugins/navsat_map/NavSatMap.qml
chapulina: Thanks, went with "Track position": c898d07

chapulina:
> I don't get this warning on focal

I can't reproduce it anymore either :smile: 

> Although I do get it for:

Ah yeah that's a pre-existing one. We should look into this when working on OOBE soon

src/plugins/navsat_map/NavSatMap.cc
```diff
@@ -0,0 +1,191 @@
...
+    this->OnRefresh();
+}
+
+/////////////////////////////////////////////////
+void NavSatMap::ProcessMessage()
+{
+  std::lock_guard<std::recursive_mutex> lock(this->dataPtr->navSatMutex);
+
+  this->newMessage(this->dataPtr->navSatMsg.latitude_deg(),
+      this->dataPtr->navSatMsg.longitude_deg());
```
jennuine, line 114: What if `latitude/longitude_deg` is not set in the new message?

jennuine (approved):

src/plugins/navsat_map/NavSatMap.cc
chapulina: Ah that's a tough one, protobuf doesn't make a distinction between `double`s that are not set or set to zero. I think it's not unreasonable to assume that these values must be set here. 

We could determine a convention like passing NaN when the fields are unknown. I see that the ROS message specifies NaNs for a missing altitude, but not for missing lat/lon :thinking: https://github.com/ros2/common_interfaces/blob/master/sensor_msgs/msg/NavSatFix.msg

jennuine: > I think it's not unreasonable to assume that these values must be set here

:+1: Sounds good to me

osrf-triage:

PR #345, opened 2022-01-06
Add project() to examples and remove hard-dependency on Ogre1

chapulina:
# 🦟 Bug fix

## Summary

This is needed to silence CMake warnings like this:

```
33: CMake Warning (dev) in CMakeLists.txt:
33:   No project() command is present.  The top-level CMakeLists.txt file must
33:   contain a literal, direct call to the project() command.  Add a line of
33:   code such as
33: 
33:     project(ProjectName)
33: 
33:   near the top of the file, but after cmake_minimum_required().
33: 
33:   CMake is pretending there is a "project(Project)" command on the first
33:   line.
33: This warning is for project developers.  Use -Wno-dev to suppress it.
```

These warnings show up on Jenkins CI since we migrated from Bionic to Focal:

* https://github.com/ignition-tooling/release-tools/pull/565

~~There's still another warning coming from `ign-rendering`, tracked here:~~

I added a workaround for the warning coming from `ign-rendering` until we fix it upstream

* https://github.com/ignitionrobotics/ign-rendering/issues/360

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #346, opened 2022-01-07
Point cloud visualization plugin

chapulina:
# 🎉 New feature

* Closes https://github.com/ignitionrobotics/ign-gazebo/issues/1156
* Requires https://github.com/ignitionrobotics/ign-msgs/pull/210 (also needs to be merged forward)

## Summary

Porting the point cloud visualization tool from [osrf/lrauv](https://github.com/osrf/lrauv/).

This needs to target Garden because it requires per-point coloring of point clouds:

* https://github.com/ignitionrobotics/ign-rendering/pull/494
* https://github.com/ignitionrobotics/ign-gui/pull/318
* https://github.com/ignitionrobotics/ign-msgs/pull/202

Features:

* Display hundreds of thousands of points with different colors
* Point positions in 3D come from a `PointCloudPacked` message
* Optionally, users can choose a separate `FloatV` topic to get values for each point
* Point size and color can be changed from the GUI
* Multiple point clouds can be displayed side-by-side, just instantiate the plugin multiple times

Known issues:

* https://github.com/ignitionrobotics/ign-rendering/issues/497

## Future work:

There are many improvements that can be made to the plugin. Here are some improvements I can think of for the future:

* Use a palette with more than 2 colors to have a richer gradient
* Optionally take color from a channel on the point cloud, instead of another topic
* Support a custom reference frame, so we can properly visualize LIDAR data - right now all points must be expressed in the world frame

## Test it

See the added example and its instructions:

pc_viz_example

Here's an examples using it to display maritime science data for LRAUV:

pc_viz

## Checklist
- [x] Added tests
- [x] ~~Updated migration guide (as needed)~~


src/plugins/point_cloud/PointCloud.hh
```diff
@@ -0,0 +1,242 @@
...
+namespace gui
+{
+namespace plugins
+{
+  class PointCloudPrivate;
+
+  /// \brief Visualize PointCloudPacked messages in the 3D scene.
+  /// By default, the whole cloud is displayed using a single color. Users
+  /// can optionally choose a topic publishing FloatV messages which will be
+  /// used to color all points with a color gradient according to their values.
```
mabelzhang, line 44: A bit more guidelines about the FloatV message should be given - its index has to match the point cloud (and what happens if it doesn't, undefined behavior or error), what kind of values (in range [0, 1] or whatever, I know it's determined automatically, but it's not clear to the reader).

src/plugins/point_cloud/PointCloud.cc
```diff
@@ -0,0 +1,496 @@
...
+  /// \brief Minimum value in latest float vector
+  public: float minFloatV{std::numeric_limits<float>::max()};
+
+  /// \brief Maximum value in latest float vector
+  public: float maxFloatV{-std::numeric_limits<float>::max()};
+
+  /// \brief Color for minimum value
+  public: ignition::math::Color minColor{1.0f, 0.0f, 0.0f, 1.0f};
+
+  /// \brief Color for maximum value
```
mabelzhang, line 86: Maybe it's beneficial to mention that these can be changed in the GUI?

mabelzhang (commented):
First pass. I looked at the code, but I need to finish rebuilding Ignition G to test this.

I know we've looked at this together a handful of times already, but I'm trying to wear a stricter hat here as it's going into Ignition, whereas in the application-specific project it's more like "anything goes."

src/plugins/point_cloud/PointCloud.hh
```diff
@@ -0,0 +1,242 @@
...
+    public: Q_INVOKABLE void Show(bool _show);
+
+    /// \brief Callback when refresh button is pressed.
+    public: Q_INVOKABLE void OnRefresh();
+
+    /// \brief Makes a request to populate the scene with markers
+    private: void PublishMarkers();
+
+    /// \brief Makes a request to delete all markers related to the point cloud.
+    private: void ClearMarkers();
```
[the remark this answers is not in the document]
chapulina: API / ABI only matters if we start installing the headers, which we aren't doing for this plugin so far. And private functions can be safely changed without breaking API / ABI. But I think that it's a good idea to move this to the private class for organization. Done.

src/plugins/point_cloud/PointCloud.hh
```diff
@@ -0,0 +1,242 @@
...
+    /// \brief Notify that topic list has changed
+    signals: void PointCloudTopicListChanged();
+
+    /// \brief Set topic to subscribe to for point cloud.
+    /// \param[in] _topicName Name of selected topic
+    public: Q_INVOKABLE void OnPointCloudTopic(const QString &_topicName);
+
+    /// \brief Callback function for float vector topic.
+    /// \param[in] _msg Float vector message
+    public: void OnFloatV(const ignition::msgs::Float_V &_msg);
```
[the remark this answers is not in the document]
chapulina: > How much does API matter in these plugins?

We're free to change it as much as we want, because the header isn't installed. All that users get is the `.so` file.

> maybe these can be privatized.

They need to be public for the QML to call them

> seem an ad-hoc data structure to represent this kind of data. 

Agreed in part. There's a reason we chose this format, right? PC fields are indeed more standard, but for the LRAUV use case, this means either:

* a single cloud that holds lots of different data, which may be wasteful if you're only interested in one type of data, 
* or multiple clouds that have XYZ duplicated.

So I think there is a valid use case for this format, even though it does feel unconventional. And I think the plugin can be extended in the future to alternatively support fields, and give users the choice of one or the other in the UI. Similarly, we should extend the plugin in the future to accept doubles, vectors, etc.

src/plugins/point_cloud/PointCloud.hh
chapulina: Added, let me know what you think.

chapulina:
Thanks for the review, @mabelzhang . I've addressed your comments and added an example. This is now open for review.

src/plugins/point_cloud/PointCloud.hh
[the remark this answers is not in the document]
mabelzhang: Yes, the floats are better than point cloud fields, but I was thinking of whether there's a data structure we hadn't thought of in the rush. In particular, I was thinking of organized point clouds with a width, height, and length. Then the ordering of the float array would be entirely up to the user to figure out, with no help from the API. In practice, that means trial and error. It's not the end of the world, it just seems not as well thought-through. It may be a corner case, as point clouds are often not organized. I won't let perfect get in the way of good.

examples/standalone/point_cloud/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+        cd <path to ign-gui>/examples/standalone/point_cloud
+
+1. Launch the example config file:
+
+        ign gui point_cloud.config
+
+3. From the build directory above, publish clouds to be visualized:
+
+    ./point_cloud
```
mabelzhang, line 40: This code block isn't showing up correctly. More spaces?

src/plugins/point_cloud/PointCloud.hh
mabelzhang: Nice! Example looks good

examples/standalone/point_cloud/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Execute Instructions
+
+1. Navigate to this directory:
+
+        cd <path to ign-gui>/examples/standalone/point_cloud
+
+1. Launch the example config file:
+
+        ign gui point_cloud.config
```
mabelzhang, line 36:
```suggestion
        ign gui -c point_cloud.config
```

mabelzhang (commented):
Thanks for adding the example! That made it a lot easier to test. Just two typos.

mabelzhang (commented):
Just pending green CI. Do we care about the doxygen warnings in the Bionic CI?

chapulina:
> Do we care about the doxygen warnings in the Bionic CI?

Ouch oh yeah we do, on it

chapulina:
> Just pending green CI.

Ok, doxygen should be fixed with edecdb8 . Besides that:

* Ubuntu's warnings should be fixed when #345 is forward-ported
* The `Helpers` test failure on Homebrew looks like a new failure unrelated to this PR

mabelzhang:
Okay re Jenkins ones.

Are these ones on the GitHub Actions intermittent?
```
  	 23 - UNIT_Publisher_TEST (Failed)
  	 35 - INTEGRATION_ExamplesBuild_TEST (Failed)
```

`UNIT_Publisher_TEST` is about X, failing only on the PR Action
```
  [GUI] [Wrn] [Application.cc:674] [QT] QXcbConnection: Could not connect to display :1.0
  [GUI] [Err] [Application.cc:678] [QT] Could not connect to any X display.
```

`INTEGRATION_ExamplesBuild_TEST` is failing on both the PR and push Actions.

chapulina:
> UNIT_Publisher_TEST is about X, failing only on the PR Action

Yup, these flaky failures due to X are all captured here:

* https://github.com/ignitionrobotics/ign-gui/issues/58

It makes life very complicated as we have to check each test and see if the failure was due to X :cry:

chapulina:
Humm now I'm confused about the `Helper` test failure on Homebrew. It doesn't fail on any of our stable branches, but I can't see why this PR would affect it.

mabelzhang:
I'm looking at the test results again, a different part of the log, and I'm seeing this related to the `ExamplesBuild` test that was failing:

```
  /github/workspace/examples/standalone/point_cloud/point_cloud.cc:31:40: error: variable 'std::atomic<bool> g_terminatePub' has initializer but incomplete type
   static std::atomic<bool> g_terminatePub(false);
                                          ^
  CMakeFiles/point_cloud.dir/build.make:62: recipe for target 'CMakeFiles/point_cloud.dir/point_cloud.cc.o' failed
  make[3]: *** [CMakeFiles/point_cloud.dir/point_cloud.cc.o] Error 1
  CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/point_cloud.dir/all' failed
  make[2]: *** [CMakeFiles/point_cloud.dir/all] Error 2
  Makefile:83: recipe for target 'all' failed
  make[1]: *** [all] Error 2
```

I think this is a real error?

chapulina:
> I think this is a real error?

Thanks for spotting, yes! On it

chapulina:
> I'm confused about the Helper test failure on Homebrew. It doesn't fail on any of our stable branches, but I can't see why this PR would affect it.

I pushed some bumpers to see if it prevents the test from segfaulting in a8ab1c0. It feels completely unrelated to this PR, but since the test consistently fails on this branch and consistently passes on `main` (search for "helper" [here](https://build.osrfoundation.org/job/ignition_gui-ci-main-homebrew-amd64/test_results_analyzer/)), I can't test if it works targeting a PR at `main` :woman_shrugging:

chapulina:
> I pushed some bumpers to see if it prevents the test from segfaulting in a8ab1c0.

It looks like they worked :pray: 

Ok, I think we're in good shape now, phew. Thanks for being thorough!

mabelzhang (approved):
Thank you for making all the fixes!

(If the bumpers aren't enough, welp, a future PR will see it :D)

scpeters:
> > I'm confused about the Helper test failure on Homebrew. It doesn't fail on any of our stable branches, but I can't see why this PR would affect it.
> 
> I pushed some bumpers to see if it prevents the test from segfaulting in [a8ab1c0](https://github.com/ignitionrobotics/ign-gui/commit/a8ab1c0322977ae118675891b71a0cd755ec501f). It feels completely unrelated to this PR, but since the test consistently fails on this branch and consistently passes on `main` (search for "helper" [here](https://build.osrfoundation.org/job/ignition_gui-ci-main-homebrew-amd64/test_results_analyzer/)), I can't test if it works targeting a PR at `main` 🤷‍♀️

sadly this test is failing on `main`, though I don't think it was caused by this PR: https://github.com/ignitionrobotics/ign-gui/issues/360

dakejahl:
>Support a custom reference frame, so we can properly visualize LIDAR data - right now all points must be expressed in the world frame

I am trying to visualize a point cloud from a camera sensor attached to a drone but the data is in body frame. Has this been implemented and can someone steer me to the right place to add this?

mabelzhang:
There should be a frame ID in the header... I would think. Line 48 in examples/standalone/point_cloud/point_cloud.cc in this PR seems to suggest it's been implemented.

There's usage example in the gz-sim integration tests e.g.
https://github.com/gazebosim/gz-sim/blob/gz-sim8/test/integration/environment_preload_system.cc
https://github.com/gazebosim/gz-sim/blob/gz-sim8/test/integration/environmental_sensor_system.cc

gz-sim environment system implementation is here
https://github.com/gazebosim/gz-sim/tree/gz-sim8/src/systems/environment_preload
https://github.com/gazebosim/gz-sim/tree/gz-sim8/src/systems/environmental_sensor_system

There's a tutorial for using the environment system, in a PR https://github.com/gazebosim/gz-sim/pull/1806

dakejahl:
@mabelzhang thanks for the reply. I took a look at those files and it's not clear to me how to access reference frame information of PointCloud data within a plugin. When I search for "frame" or "model" in the **src/plugins/** I don't see any other examples of converting between reference frames for sensor data attached to a moving model. The goal is to visualize obstacle detect and avoid on PX4. Thanks for the help!

Here's the model I'm using for the drone, you can see how the camera is attached
https://github.com/PX4/PX4-Autopilot/blob/main/Tools/simulation/gz/models/x500_depth/model.sdf

I have some example code where I am downsampling a camera point cloud and republishing it
https://github.com/dakejahl/PX4-Autopilot/commit/02ef5280a647ffc83d8de74f0830476f1fe3d02a

Here's a pic of a drone with a point cloud looking at a block some distance from the origin
image

mabelzhang:
Hey, okay, so first, we really shouldn't be starting a new topic in a merged PR. The more correct thing to do would be to open an issue and tag this PR as the relevant source code. I replied with links to the source code because they depend on code in this PR. But with the new details you posted, it should really be in a new issue ticket for better visibility.

Please keep that in mind and open a new ticket if you post new information.

@arjo129 can people set frame ID when visualizing point clouds? That would be part of the environment system right?

arjo129:
To be fair this has nothing to do with the environment system. We should open a ticket, I suspect point clouds don't follow frame_ids set :scream: .

dakejahl:
Sorry about that! I figured posting here was the best way to get the attention of the person who knows! :grin:

Ticket here
https://github.com/gazebosim/gz-gui/issues/595

PR #349, opened 2022-01-14
Improve inserting WorldControl and WorldStats from menu

chapulina:
# 🦟 Bug fix

* Part of https://github.com/ignitionrobotics/ign-gazebo/issues/1254

## Summary

Currently, if `WorldControl` or `WorldStats` are inserted from the menu, they don't display anything and are therefore unusable. That's because all their elements are optional and hidden by default (which we should consider changing).

This PR works around that by displaying all widgets if no elements are passed onto the plugin. The assumption is that the user would never want to hide all elements at the same time.

With this PR, these plugins can be inserted from the menu and they work with Gazebo:

world_gui_from_menu

## Checklist
- [x] Added tests
- [x] ~~Updated migration guide (as needed)~~


src/plugins/world_control/WorldControl.hh
```diff
@@ -54,15 +54,19 @@ namespace plugins
   ///
   /// ## Configuration
   ///
-  /// * \<play_pause\> : Set to true to see a play/pause and step buttons,
+  /// * \<play_pause\> : Set to true to see a play/pause button,
   ///                    false by default.
+  /// * \<step\> : Set to true to see a step button, false by default.
```
jennuine, line 59: Should we update the defaults be `true` as well?

jennuine (approved):
LGTM, left a minor comment

src/plugins/world_control/WorldControl.hh
chapulina: I tried to do this in a way that the defaults wouldn't need to be changed, because that would break existing behaviour. For example, a user who currently has only `<play_pause>` will continue not getting the step button, because `<step>` is false by default.

The only situation that changed is when the `<plugin>` tag is completely empty and has the effect of flipping all defaults to true. This is technically a change in behaviour, but I think that wouldn't break any existing users, because no one should be using the plugin completely empty, and it would fix a usability bug.

Let me know if the reasoning sounds ok to you

jennuine: Ah ok, thanks for explaining. Sounds good to me

osrf-triage:

PR #359, opened 2022-02-22
Removed unused code in MinimalScene plugin

### LINKED ISSUE #358: What is the planned relationship between MinimumScene and TransportSceneManager?
I noticed that MinimumScene has the following, but then I came across TransportSceneManager which seems to manage the same set of topics and service. Is the plan that this functionality will be removed from MinimumScene and TransportSceneManager should be used going forward or is there another intended purpose of TransportSceneManager?

``` c++
    /// \brief Scene service. If not empty, a request will be made to get the
    /// scene information using this service and the renderer will populate the
    /// scene based on the response data
    public: std::string sceneService;

    /// \brief Scene pose topic. If not empty, a node will subcribe to this
    /// topic to get pose updates of objects in the scene
    public: std::string poseTopic;

    /// \brief Ign-transport deletion topic name
    public: std::string deletionTopic;

    /// \brief Ign-transport scene topic name
    /// New scene messages will be published to this topic when an entities are
    /// added
    public: std::string sceneTopic;
```

ahcorde:

# 🦟 Bug fix

Fixes #358

## Summary

This code is not used and it should be part of the TransportSceneManager

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.hh
+++ b/src/plugins/minimal_scene/MinimalScene.hh
@@ -205,23 +205,6 @@ namespace plugins
     /// \brief Flag to indicate texture size has changed.
     public: bool textureDirty = false;
 
-    /// \brief Scene service. If not empty, a request will be made to get the
-    /// scene information using this service and the renderer will populate the
-    /// scene based on the response data
-    public: std::string sceneService;
-
-    /// \brief Scene pose topic. If not empty, a node will subcribe to this
-    /// topic to get pose updates of objects in the scene
-    public: std::string poseTopic;
-
-    /// \brief Ign-transport deletion topic name
-    public: std::string deletionTopic;
-
-    /// \brief Ign-transport scene topic name
-    /// New scene messages will be published to this topic when an entities are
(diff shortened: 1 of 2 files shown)


src/plugins/minimal_scene/MinimalScene.hh
```diff
@@ -205,23 +205,6 @@ namespace plugins
     /// \brief Flag to indicate texture size has changed.
     public: bool textureDirty = false;
 
-    /// \brief Scene service. If not empty, a request will be made to get the
-    /// scene information using this service and the renderer will populate the
-    /// scene based on the response data
-    public: std::string sceneService;
```
jennuine, line 211: Removing data members and functions doesn't break abi? Is this because it's a plugin?

chapulina: Yeah we're not installing this header file, so we don't need to worry about ABI

chapulina (approved):
LGTM, we just overlooked these when splitting the plugin.

PR #369, opened 2022-03-07
Added Snackbar qtquick object

ahcorde:
# 🎉 New feature

## Summary

Create a IgnSnackBar QtQuick object to show message to the user in the same manner by plugins

It could be :
 - persistant and then you need to click inside the GUI
 - or with a timeout

snackback

Added snackbar dialog to:
 - Save configuration
 - Teleop topic
 - Snapshot
 - Image view update

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gui/MainWindow.hh
+++ b/include/ignition/gui/MainWindow.hh
@@ -569,6 +569,9 @@ namespace ignition
       signals: void notify(const QString &_message);
 
       /// \brief Displays a message to the user
+      /// \param[in] _message Message to show
+      /// \param[in] _duration Time in milliseconds that the message will
+      /// appear
       signals: void notifyWithDuration(const QString &_message, int _duration);
 
       /// \internal
--- a/include/ignition/gui/MainWindow.hh
+++ b/include/ignition/gui/MainWindow.hh
@@ -568,6 +568,9 @@ namespace ignition
       /// \brief Displays a message to the user
       signals: void notify(const QString &_message);
 
+      /// \brief Displays a message to the user
+      signals: void notifyWithDuration(const QString &_message, int _duration);
+
       /// \internal
       /// \brief Private data pointer
       private: std::unique_ptr<MainWindowPrivate> dataPtr;
--- a/include/ignition/gui/qml/IgnSnackBar.qml
+++ b/include/ignition/gui/qml/IgnSnackBar.qml
@@ -15,6 +15,7 @@
  *
 */
 
+import QtGraphicalEffects 1.0
 import QtQuick 2.9
 import QtQuick.Controls 2.2
 import QtQuick.Controls.Material 2.2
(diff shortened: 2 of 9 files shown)


include/ignition/gui/MainWindow.hh
```diff
@@ -568,6 +568,9 @@ namespace ignition
       /// \brief Displays a message to the user
       signals: void notify(const QString &_message);
 
+      /// \brief Displays a message to the user
+      signals: void notifyWithDuration(const QString &_message, int _duration);
```
chapulina, line 572: Missing `param` docs

include/ignition/gui/qml/IgnSnackBar.qml
```diff
@@ -0,0 +1,79 @@
...
+  modal: duration == 0
+  focus: duration == 0
+  x: (window.width - width) / 2
+  y: window.height - window.height / 6
+  width: window.width - window.width / 6
+  contentHeight: notificationColumn.height
+
+  background: Rectangle {
+    border.color: "#444"
+    color: "#c9daf8"
```
chapulina, line 35: We should avoid using custom colors as much as possible and instead use colors from the current palette (`Material.primary`, `Material.accent`, `Material.background`). 

The [official spec](https://material.io/components/snackbars#anatomy) recommends a grey color, but I think QtQuick doesn't make it straight-forward to have that follow a dark / light theme. So I think we could stick to `Material.background` here.

src/plugins/teleop/Teleop.cc
```diff
@@ -140,8 +140,19 @@ void Teleop::OnTopicSelection(const QString &_topic)
...
+    App()->findChild<MainWindow *>()->notifyWithDuration(
+      QString::fromStdString("Error when advertising topic: " +
+        this->dataPtr->topic), 1000);
+    ignerr << "Error when advertising topic: " <<
+      this->dataPtr->topic << std::endl;
+  }
+  else
+  {
+    App()->findChild<MainWindow *>()->notifyWithDuration(
+      QString::fromStdString("A new topic has been entered: '" +
```
chapulina, line 153: How about:

```suggestion
      QString::fromStdString("Subscribing to topic: '" +
```

include/ignition/gui/qml/IgnSnackBar.qml
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // Duration of the snackbar. If duration is equal to zero then
+  // you should click somewhere in Ignition Gazebo to close it.
+  property int duration: 1000
```
chapulina, line 40: I'm finding 1 second too short to read some of the snackbars on this PR. The Material spec [recommends](https://material.io/components/snackbars#behavior) "_They automatically disappear from the screen after a minimum of four seconds, and a maximum of ten seconds._", so how about defaulting to `4000` and using that throughout this PR?

chapulina (changes requested):
+1 to the centralized approach using a signal from the `MainWindow` :+1: This should make it easier to improve the snackbar in the future. For example, in the future, if we start getting too many notifications at once, we may want to implement some logic to display them consecutively.

chapulina (approved):
LGTM! I just pushed a small change in eb58ef3aa3f05a6a379bfc2cb3052e709e658351 to change the border to a shadow.

Before:

border

After:

shadow

osrf-triage:

https://community.gazebosim.org/t/new-ignition-releases-2022-04-13-fortress-edifice/1367/1

PR #370, opened 2022-03-08
Added array to snackbar qml

ahcorde:
# 🎉 New feature

## Summary

if we start getting too many notifications at once to show in the snackbar, this logic will  display them consecutively.

ESC will close the popup, otherwise with the mouse click some of the popup will close inmediately if you are clicking several times to configure something.

Build on top of https://github.com/ignitionrobotics/ign-gui/pull/369

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gui/qml/IgnSnackBar.qml
+++ b/include/ignition/gui/qml/IgnSnackBar.qml
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+import QtQuick 2.9
+import QtQuick.Controls 2.2
+import QtQuick.Layouts 1.3
+import QtQuick.Window 2.2
+import QtQuick.Dialogs 1.0
+
+Popup {
+  id: snackbar
+  modal: duration == 0
+  focus: duration == 0
+  x: (window.width - width) / 2
+  y: window.height - window.height / 6
+  width: window.width - window.width / 6
+  contentHeight: notificationColumn.height
+
+  background: Rectangle {
+    border.color: "#444"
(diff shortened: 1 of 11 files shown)


include/ignition/gui/qml/IgnSnackBar.qml
```diff
@@ -20,16 +20,24 @@ import QtQuick.Controls 2.2
...
 Popup {
   id: snackbar
-  modal: duration == 0
-  focus: duration == 0
   x: (window.width - width) / 2
   y: window.height - window.height / 6
   width: window.width - window.width / 6
   contentHeight: notificationColumn.height
 
+  closePolicy: Popup.CloseOnEscape
```
chapulina, line 32: > ESC will close the popup, otherwise with the mouse click some of the popup will close inmediately if you are clicking several times to configure something.

I understand that this is meant to protect the user from closing the popup by mistake, but clicking outside it to close it is so common that I think users may be confused when that doesn't work (I was when I first tried this PR).

For popups that we really don't want the user to miss, how about supporting an optional "Dismiss" button, so the snackbar can only be dismissed by clicking that?

ahcorde:
@chapulina 

 - Added a Dismiss button 
 - Aligned text
 - used bold to highlight some relevant information such as: topics or paths

chapulina:
> Added a Dismiss button

Thanks! I was thinking of making that optional though. Most implementations have buttons as optional, see

* https://www.webcomponents.org/element/@polymer/paper-toast/demo/demo/index.html
* https://material.angular.io/components/snack-bar/overview

Usually the button text is configurable and there's a way to register a callback when the button is clicked. I don't think we need all these features just yet, but it would be good to:

* structure the API in a way that it can be extended later
* the button should be optional for now, so messages that aren't very important have a timeout and can be dismissed by the usual means (like subscription to topics), but messages that are important have a button that must be clicked to dismiss (like a failure to save a file).
* make the button flat like in the implementations above (from the spec "_Don’t use a filled or elevated button in a snackbar, as it draws too much attention._")

> Aligned text

I think it would be better to follow the [spec](https://material.io/components/snackbars#specs) and align the text to the left like it was already.

chapulina:
@ahcorde , I have some suggestions on #375, let me know what you think

chapulina:
Argh, Bionic is broken with

```
   qrc:/qml/IgnSnackBar.qml:48 Cannot assign to non-existent property "horizontalPadding"
```

That's something I added on #375. Let me find an alternative

chapulina (approved):
LGTM once CI comes back ok

chapulina:
@osrf-jenkins run tests please

chapulina:
Bionic is happy now, the webhooks have been acting up all day. Merging.

osrf-triage:

https://community.gazebosim.org/t/new-ignition-releases-2022-04-13-fortress-edifice/1367/1

PR #385, opened 2022-04-13
Bumps in garden : ign-gui7

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/685

### DIFF
--- a/src/plugins/world_control/WorldControl.cc
+++ b/src/plugins/world_control/WorldControl.cc
@@ -64,7 +64,7 @@ namespace plugins
 
     /// \brief Whether server communication should occur through an event (true)
     /// or service (false). The service option is used by default for
-    /// ign-gui7, and should be changed to use the event by default in ign-gui7.
+    /// ign-gui6, and should be changed to use the event by default in ign-gui7.
     public: bool useEvent{false};
   };
 }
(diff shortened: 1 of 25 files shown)


src/plugins/world_control/WorldControl.cc
```diff
@@ -64,7 +64,7 @@ namespace plugins
 
     /// \brief Whether server communication should occur through an event (true)
     /// or service (false). The service option is used by default for
-    /// ign-gui6, and should be changed to use the event by default in ign-gui7.
+    /// ign-gui7, and should be changed to use the event by default in ign-gui7.
```
chapulina, line 67: Oops

chapulina (approved):
Just a minor greedy match. LGTM otherwise, once it has 🟢 CI

src/plugins/world_control/WorldControl.cc
methylDragon: https://github.com/ignitionrobotics/ign-gui/pull/385/commits/a007fef08d52bec8a701f1737e023d03d9009cf5

chapulina:
The new test failures are caused by a bad combination of `ign-cmake2` and `ign-cmake3` and their `utilities` package. I'm going to merge this as is because:

* We have various merged and open PRs to get rid of `utilities`
* The next wave of bumps should remove the double `ign-cmake2` and `ign-cmake3`

Let's not block on this, merging!

PR #392, opened 2022-04-30
Bash completion for flags

mabelzhang:
# 🎉 New feature

Part of https://github.com/ignitionrobotics/ign-tools/issues/1
Used together with https://github.com/ignitionrobotics/ign-tools/pull/87

## Summary

Bash completion script for flags available to this subcommand.
This is a standalone function that depends on a new script in https://github.com/ignitionrobotics/ign-tools/pull/87 to call this.

## Test it

After sourcing the dependent script in ign-tools, source the script in this PR:
```
. src/cmd/gui.bash_completion.sh
```

If you just tab, you get the default system tab-completion that shows the files in the current directory:
```
$ ign gui <tab>
```

If you type `-`, and then tab, you get the tab-completion for all the flags available for this subcommand:
```
$ ign gui -<tab>
--config         --help           --standalone     --versions       -h               -s               
--force-version  --list           --verbose        -c               -l               -v               
```

Then you can type partial flags, and it'll do tab-completion normally for you, such as
```
$ ign gui --ver
--verbose   --versions  
```

It also works for multiple flags, like `ign gui --<flag1> --<flag2> ...`

## Checklist
- [ ] Added tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,9 +69,11 @@ set(IGN_MSGS_VER ${ignition-msgs5_VERSION_MAJOR})
 
 #--------------------------------------
 # Find if ign command is available
-find_program(HAVE_IGN_TOOLS ign)
-if (HAVE_IGN_TOOLS)
-  set(IGN_TOOLS_VER 1)
+
+ign_find_package(ignition-tools QUIET)
+if (ignition-tools_FOUND)
+  set (HAVE_IGN_TOOLS TRUE)
+  set (IGN_TOOLS_VER 1)
 endif()
 
 #--------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,13 +68,11 @@ ign_find_package(ignition-msgs5 REQUIRED)
 set(IGN_MSGS_VER ${ignition-msgs5_VERSION_MAJOR})
 
 #--------------------------------------
-# Find if ign command is available
-
-ign_find_package(ignition-tools QUIET)
-if (ignition-tools_FOUND)
-  set (HAVE_IGN_TOOLS TRUE)
-  set (IGN_TOOLS_VER 1)
-endif()
+# Find if command is available. This is used to enable tests.
+# Note that CLI files are installed regardless of whether the dependency is
+# available during build time
+find_program(HAVE_IGN_TOOLS ign)
+set (IGN_TOOLS_VER 1)
 
 #--------------------------------------
 # Find QT
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,9 @@ set(IGN_MSGS_VER ${ignition-msgs5_VERSION_MAJOR})
 #--------------------------------------
 # Find if ign command is available
 find_program(HAVE_IGN_TOOLS ign)
+if (HAVE_IGN_TOOLS)
+  set(IGN_TOOLS_VER 1)
+endif()
 
 #--------------------------------------
 # Find QT
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,7 +52,7 @@ ign_build_tests(TYPE UNIT
 
 foreach(test ${gtest_targets})
   target_compile_definitions(${test} PRIVATE
-    "IGN_GUI_SOURCE_DIR=\"${PROJECT_SOURCE_DIR}\"")
+    "GZ_GUI_SOURCE_DIR=\"${PROJECT_SOURCE_DIR}\"")
 endforeach()
 
 add_subdirectory(cmd)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -46,7 +46,14 @@ ign_build_tests(TYPE UNIT
                 LIB_DEPS
                   ${IGNITION-MATH_LIBRARIES}
                   TINYXML2::TINYXML2
+                TEST_LIST
+                  gtest_targets
 )
+
+foreach(test ${gtest_targets})
+  target_compile_definitions(${test} PRIVATE
+    "IGN_GUI_SOURCE_DIR=\"${PROJECT_SOURCE_DIR}\"")
+endforeach()
 
 add_subdirectory(cmd)
 add_subdirectory(plugins)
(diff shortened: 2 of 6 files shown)


chapulina (commented):
Same comments as in `plugin`, let's use `gz` wherever we can, and see if it's possible to write a test that checks that the flags are the same on the autocomplete file and the command itself.

mabelzhang:
I think the comments are addressed. Ready for another pass.

src/cmd/gui.bash_completion.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
ahcorde, line 1: LICENSE ?

ahcorde (changes requested):

src/CMakeLists.txt
```diff
@@ -46,7 +46,14 @@ ign_build_tests(TYPE UNIT
                 LIB_DEPS
                   ${IGNITION-MATH_LIBRARIES}
                   TINYXML2::TINYXML2
+                TEST_LIST
+                  gtest_targets
 )
 
+foreach(test ${gtest_targets})
+  target_compile_definitions(${test} PRIVATE
+    "IGN_GUI_SOURCE_DIR=\"${PROJECT_SOURCE_DIR}\"")
```
chapulina, line 55:
```suggestion
    "GZ_GUI_SOURCE_DIR=\"${PROJECT_SOURCE_DIR}\"")
```

src/cmd/gui.bash_completion.sh
mabelzhang: 5944868b

src/CMakeLists.txt
mabelzhang: 5944868b

ahcorde (approved):

src/ign_TEST.cc
```diff
@@ -96,3 +97,33 @@ TEST_F(CmdLine, IGN_UTILS_TEST_ENABLED_ONLY_ON_LINUX(list))
   EXPECT_TRUE(common::exists(common::joinPaths(this->kFakeHome, ".ignition",
       "gui")));
 }
+
+//////////////////////////////////////////////////
+/// \brief Check --help message and bash completion script for consistent flags
+TEST(ignTest, GuiHelpVsCompletionFlags)
```
chapulina, line 103: The only concern I have with the way this test is currently structured is that it relies on developers remembering to update it whenever they add a new flag. And at this point, they may as well remember to update the bash completion as well.

Ideally the test wouldn't have the options hardcoded, so that whenever a new option is added, the test automatically detects that it's missing from bash completion.

mabelzhang: I'll add that the tests themselves are not foolproof, especially for the single letter flags. Strings like `-a` could match many things in a bash script file.

To automate the tests, I guess it could be possible to add a function to the bash scripts returning the word list, and the test can source the bash script and then call that function. It probably needs to be a separate function than the existing one that's used by the actual bash completion, so the word list might need to be a global variable.

mabelzhang: [6f7291b](https://github.com/gazebosim/gz-gui/pull/392/commits/6f7291bf3b06f9c71f5aa3eff7fc3c919b69209d) changes test so that it's no longer hardcoded. Flags are now in a variable in the bash script, and a new bash script function returns them. The test calls this function.

chapulina (changes requested):

src/ign_TEST.cc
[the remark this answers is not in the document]
mabelzhang: That's weird this file is being compiled for Apple in the first place, given this block

https://github.com/gazebosim/gz-gui/blob/5cd426d2ce637ab7504839e8d1d9878488bfa780/src/CMakeLists.txt#L37-L41

e7211eba attempts to fix the Apple test... let's see.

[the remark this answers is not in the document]
mabelzhang: bab2541 seems to have fixed the Ubuntu test. Now `GuiHelpVsCompletionFlags` passes on Bionic and Focal. Homebrew still running.

CMakeLists.txt
```diff
@@ -69,7 +69,12 @@ set(IGN_MSGS_VER ${ignition-msgs5_VERSION_MAJOR})
 
 #--------------------------------------
 # Find if ign command is available
-find_program(HAVE_IGN_TOOLS ign)
+
+ign_find_package(ignition-tools QUIET)
```
chapulina, line 73: This is reverting https://github.com/gazebosim/gz-gui/pull/240. My understanding is that the CMake / pkg-config module is being removed from `ign-tools` (see https://github.com/gazebo-tooling/release-tools/issues/472), so this won't work anymore.

chapulina (changes requested):

CMakeLists.txt
chapulina: Reverted in [fe2be11](https://github.com/gazebosim/gz-gui/pull/392/commits/fe2be11e57bf2c3cc2b838f9d5144167a43e3aef) and added a note to avoid future confusion.

chapulina (approved):
Thanks for iterating! It's working for me after I source:

```
. install/share/gz/gz1.completion
```

chapulina:
Fixed the tests in [2866b7a](https://github.com/gazebosim/gz-gui/pull/392/commits/2866b7a9ea535d8b6dafd854d0a58827df0a5776), inspired by https://github.com/gazebosim/gz-plugin/pull/57

Blast545:
@mabelzhang This PR introduced a regression in the Gazebo Citadel Windows buildfarm, see:
[ign_gui-ign-3-win#93/testReport/junit/(root)/ignTest/GuiHelpVsCompletionFlags/](https://build.osrfoundation.org/job/ign_gui-ign-3-win/93/testReport/junit/(root)/ignTest/GuiHelpVsCompletionFlags/)

```
C:\Jenkins\workspace\ign_gui-ign-3-win\ws\ign-gui\src\ign_TEST.cc:131
Expected: (std::string::npos) != (helpOutput.find(flag)), actual: 18446744073709551615 vs 18446744073709551615
'HAVE_IGN_TOOLS-NOTFOUND' is not recognized as an internal or external command,
operable program or batch file.
```

Would you please take a look? It seems like an extra environment variable has to be set on Windows to get the proper output from the command line on that test.

chapulina:
> This PR introduced a regression in the Gazebo Citadel Windows

Ah sorry that's on me, I totally missed the failing Windows test. On it.

PR #405, opened 2022-05-24
Shortcut to search and auto highlight search bar

AzulRadio:
Signed-off-by: youhy <haoyuan2019@outlook.com>

# New feature(s)

## Summary
There are 2 features added in this PR. Both of them are small and are about the search bar/menu so I combine them into one PR.

Feature 1. When we close the search bar for adding plugins, the text in it will be preserved. If we want to add another plugin, we need to delete the first one. That's a little annoying. Now after adding a plugin, text in the search bar will be auto highlighted so the search bar can be cleared easier. 

No highlight if no plugin is added.

demo_better_highligh

Feature 2. Press `/` to open the plugin search bar

demo_shortcut

## Test it
Try it

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gui/qml/Main.qml
+++ b/include/ignition/gui/qml/Main.qml
@@ -151,6 +151,11 @@ ApplicationWindow
     onActivated: close()
   }
 
+  Shortcut {
+    sequence: "/"
+    onActivated: pluginMenu.open()
+  }
+
   /**
    * Top toolbar
    */
--- a/include/ignition/gui/qml/PluginMenu.qml
+++ b/include/ignition/gui/qml/PluginMenu.qml
@@ -38,14 +38,10 @@ Popup {
       Material.color(Material.Grey, Material.Shade200):
       Material.color(Material.Grey, Material.Shade900);
 
-  property bool pluginAdded: false
 
   onOpened: {
     searchField.forceActiveFocus()
-    if (pluginAdded) {
-      searchField.selectAll()
-      pluginAdded = false;
-    }
+    searchField.selectAll()
   }
 
   ColumnLayout {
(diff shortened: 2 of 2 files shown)


include/ignition/gui/qml/Main.qml
```diff
@@ -151,6 +151,11 @@ ApplicationWindow
     onActivated: close()
   }
 
+  Shortcut {
+    sequence: "/"
```
jennuine, line 155: Initially I thought this may be a problem for when we're typing in topic names (e.g., in the topic echo plugin) but it doesn't seem to be the case. Can you verify this @AzulRadio? Go to the plugin menu > topic echo > then in the text field type `/gui/camera/pose` and make sure that the plugin menu doesn't open during that time

include/ignition/gui/qml/PluginMenu.qml
```diff
@@ -38,7 +38,15 @@ Popup {
...
       Material.color(Material.Grey, Material.Shade900);
 
-  onOpened: searchField.forceActiveFocus()
+  property bool pluginAdded: false
+
+  onOpened: {
+    searchField.forceActiveFocus()
+    if (pluginAdded) {
+      searchField.selectAll()
+      pluginAdded = false;
```
jennuine, line 47: IMO I think we should always have it highlighted when the menu is opened. If users want to continue typing where they left off they can use the :arrow_right: key

AzulRadio: Changed back in aaa49a59

include/ignition/gui/qml/Main.qml
jennuine: Great, can you add the new shortcut to https://github.com/gazebosim/docs/blob/master/fortress/hotkeys.md please?

jennuine (approved):

include/ignition/gui/qml/Main.qml
AzulRadio: Done in docs#250

PR #406, opened 2022-05-25
ign -> gz Migrate Examples : gz-gui

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/711
Related to: https://github.com/gazebosim/gz-gui/pull/401

### DIFF
--- a/examples/plugin/gz_components/CMakeLists.txt
+++ b/examples/plugin/gz_components/CMakeLists.txt
@@ -21,7 +21,7 @@ find_package(ignition-common5 REQUIRED)
 
 set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${IGNITION-GUI_CXX_FLAGS}")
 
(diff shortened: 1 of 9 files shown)


src/Application.cc
```diff
@@ -91,15 +91,15 @@ using namespace gui;
 Application::Application(int &_argc, char **_argv, const WindowType _type)
   : QApplication(_argc, _argv), dataPtr(new ApplicationPrivate)
 {
-  igndbg << "Initializing application." << std::endl;
+  gzdbg << "Initializing application." << std::endl;
```
chapulina, line 94: Oops, these can't go in yet

chapulina (changes requested):
The moved files LGTM, but we have to revert the `gzdbg` changes

src/Application.cc
methylDragon: Oops

methylDragon: Force pushed and now this commit is :boom:

chapulina (approved):

PR #413, opened 2022-06-06
ign -> gz Upstream Macro Migration : gz-gui

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/737

### DIFF
--- a/src/Plugin.cc
+++ b/src/Plugin.cc
@@ -181,10 +181,7 @@ void Plugin::Load(const tinyxml2::XMLElement *_pluginElem)
     {
       gzwarn << "The `ignition-gui` element is deprecated. Please use "
       << "`gz-gui` instead." << std::endl;
-    }
-    else
-    {
-      this->LoadCommonConfig(_pluginElem->FirstChildElement("gz-gui"));
+      this->LoadCommonConfig(_pluginElem->FirstChildElement("ignition-gui"));
     }
   }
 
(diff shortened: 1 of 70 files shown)


chapulina (approved):
LGTM once upstream PRs are merged and CI is happy 🟢

src/Plugin.cc
```diff
@@ -167,26 +167,45 @@ void Plugin::Load(const tinyxml2::XMLElement *_pluginElem)
...
+  {
+    guiElem = _pluginElem->FirstChildElement("ignition-gui");
+    if (guiElem)
+    {
+      gzwarn << "The `ignition-gui` element is deprecated. Please use "
+      << "`gz-gui` instead." << std::endl;
+    }
+    else
+    {
+      this->LoadCommonConfig(_pluginElem->FirstChildElement("gz-gui"));
```
chapulina, line 187: Should be up in the `if` together with the warning, and say `ignition-gui`?

src/Plugin.cc
```diff
@@ -290,24 +309,35 @@ std::string Plugin::ConfigStr()
...
-  // <ignition-gui>
-  auto ignGuiElem = pluginElem->FirstChildElement("ignition-gui");
-  if (!ignGuiElem)
+  // <gz-gui>
+  auto guiElem = pluginElem->FirstChildElement("gz-gui");
+  if (!guiElem)
   {
-    ignGuiElem = doc.NewElement("ignition-gui");
-    pluginElem->InsertEndChild(ignGuiElem);
+    // TODO(CH3): Deprecated. Remove on tock.
```
chapulina, line 320: This function is used for saving the config, so I think just changing `ignition` to `gz` is enough. No need to tick-tock. All new saved configs will have `gz`, which is ok. The tick-tock you did for loading old configs that contain `ignition` is enough to tick-tock this feature.

chapulina (changes requested):

src/Plugin.cc
methylDragon: https://github.com/gazebosim/gz-gui/pull/413/commits/fe598592f38834a889c2d6bde41bcaee30b3c1d5 un-tick-tocked!

chapulina:

PR #417, opened 2022-06-12
ign -> gz CLI Migration : gz-gui

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/739


methylDragon:
CI failure needs the merge of `gz-tools` so the `gz` executable is available

methylDragon:

chapulina (changes requested):
Ahh `include/ignition/gui/ign.hh` is installed, so we shouldn't move it. I think it's ok to leave it as it is, because it's already deprecated inside the `ignition` folder, and the new `include/gz/gui/gz.hh` header is already correct.

methylDragon:
> Ahh `include/ignition/gui/ign.hh` is installed, so we shouldn't move it. I think it's ok to leave it as it is, because it's already deprecated inside the `ignition` folder, and the new `include/gz/gui/gz.hh` header is already correct.

Fixed! Move no longer occurs in https://github.com/gazebosim/gz-gui/pull/417/commits/a2489b397fbfdc499a3ab9fc5179b8ab78f9e2aa

chapulina (approved):
`UNIT_gz_TEST` passes on Linux and macOS!

PR #418, opened 2022-06-14
Allow Dialogs to have a MainWindow independent config

mayman99:
Signed-off-by: Mohamad <mohamad@openrobotics.org>
🎉 New feature 

This PR allows Dialogs to read and write to the config file.
The motivation for this PR is, all the current reading/writing of the `Config` is implemented in `MainWindow`, and the Application has not been initialized yet then there is not `MainWindow` object.

a dialog config looks like `<dialog name=\"dialog_name\" attribute_1=\"true\"/>"`

Workflow:

1. Set dialog name `SetName`
2. set the dialog default config using `SetDefaultConfig`
3. Read an attribute of the dialog config using `ReadAttribute`
4. if the dialog config doesn't exist in the existing config file
    5. append the dialog default config to the config file
    6. read from the written config
7.  if the dialog config exists in the existing config file, read it and return the value as string
8. update an attribute value in the dialog config using `WriteAttributre`

The changes here are used in [this PR](https://github.com/gazebosim/gz-sim/pull/1536).
## Checklist
- [x] Added tests


chapulina (commented):
> The motivation for this PR is, all the current reading/writing of the Config is implemented in MainWindow, and the Application has not been initialized yet then there is not MainWindow object.

I'm not sure I understand this part. The application has to be initialized before the dialog. Can't we use leave the config path handling to the `Application`? It would be nice if we didn't need another config file and could reuse the same one. The `<dialog>` elements could be siblings of the `<window>` element.

---

Have you seen the `WindowConfig` struct? That's used to hold an in-memory representation of an XML config file, load from a file and write to it. It would be interesting to follow the same pattern for consistency of the API, creating a new `DialogConfig` struct.

mayman99:
Turning the Dialog class to be more like the MainWindow, like to have a similar API is better than coming up with a new API like how I just did on the PR.
I'll be modifying it to have a similar `DialogConfig` struct and a similar API, thanks for the suggestion

mayman99:
Updated the PR per VC, except of using templates with `ReadAttributeValue`, as it results in a seg-fault when using in `gazebo`, which needs further debugging to figure out why.

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+  // Write config file
+  tinyxml2::XMLPrinter printer;
+  doc.Print(&printer);
+
+  std::string config = printer.CStr();
+  std::ofstream out(_path.c_str(), std::ios::out);
+  if (!out)
+  {
+    std::string str = "Unable to open file: " + _path;
+    str += ".\nCheck file permissions.";
```
chapulina, line 124: The `str` variable isn't used anywhere, did you mean to print an error?

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+      missingDoc.Print(&defaultPrinter);
+    }
+  }
+
+  // Write config file
+  tinyxml2::XMLPrinter printer;
+  doc.Print(&printer);
+
+  // Don't write the xml version decleration if file exists
+  if (configExists){
```
chapulina, line 211: Style nit

```suggestion
  if (configExists)
  {
```

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+  igndbg << "Setting Quick start dialog default config." << std::endl;
+
+  config += printer.CStr();
+  config += defaultPrinter.CStr();
+  std::ofstream out(_path.c_str(), std::ios::out);
+  if (!out)
+  {
+    std::string str = "Unable to open file: " + _path;
+    str += ".\nCheck file permissions.";
+    igndbg << str << std::endl;
```
chapulina, line 223: This looks like an error

```suggestion
    ignerr << str << std::endl;
```

You could also save a copy and just stream the error directly to `ignerr` without going through an intermediate variable `str`

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+    configExists = false;
+    doc.Parse(this->dataPtr->config.c_str());
+    // Process each dialog
+    for (auto dialogElem = doc.FirstChildElement("dialog");
+      dialogElem != nullptr;
+      dialogElem = dialogElem->NextSiblingElement("dialog"))
+    {
+      if(dialogElem->Attribute("name") == this->objectName().toStdString())
+      {
+        value = dialogElem->Attribute(_attribute.c_str());
```
chapulina, line 162: Currently, if there are multiple dialog elements with the same name, the function will return the last of them. I think that behaviour is ok. Can you document it?

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+    {
+      return "";
+    }
+
+    // Process each existing dialog
+    for (auto dialogElem = doc.FirstChildElement("dialog");
+      dialogElem != nullptr;
+      dialogElem = dialogElem->NextSiblingElement("dialog"))
+    {
+      if(dialogElem->Attribute("name") == this->objectName().toStdString())
```
chapulina, line 179:
```suggestion
      if (dialogElem->Attribute("name") == this->objectName().toStdString())
```

src/Dialog_TEST.cc
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
chapulina, line 2:
```suggestion
 * Copyright (C) 2022 Open Source Robotics Foundation
```

src/Dialog_TEST.cc
```diff
@@ -0,0 +1,93 @@
...
+      "allow");
+    EXPECT_EQ(allow, "");
+  }
+
+  // Read a non existing attribute
+  {
+    dialog->setObjectName("quick_menu");
+    dialog->SetDefaultConfig(std::string(
+      "<dialog name=\"quick_menu\" show=\"true\"/>"));
+    std::string allow = dialog->ReadAttributeValue(app.DefaultConfigPath(),
```
chapulina, line 64: You could check that the config file didn't exist, and then it was created by this function

src/Dialog.cc
```diff
@@ -75,3 +78,153 @@ QQuickItem *Dialog::RootItem() const
...
+  else
+  {
+    auto success = !doc.LoadFile(_path.c_str());
+    if (!success)
+    {
+      return "";
+    }
+
+    // Process each existing dialog
+    for (auto dialogElem = doc.FirstChildElement("dialog");
```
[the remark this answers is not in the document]
mayman99: maybe I could use a lambda to hold the loop :thinking:

include/ignition/gui/Dialog.hh
```diff
@@ -55,6 +56,28 @@ namespace ignition
...
+      /// \param[in] _path config path
+      /// \param[in] _attribute XMLElement attribute name
+      /// \param[in] _value XMLElement attribute value
+      /// \return true if written to config file
+      public: bool UpdateConfigAttribute(
+        const std::string &_path, const std::string &_attribute,
+        const bool _value) const;
+
+      /// \brief Gets a config attribute value, if not found in config
+      /// \brief write the default in the config and get it.
```
chapulina, line 73:
```suggestion
      /// write the default in the config and get it.
```

This is a continuation of the brief above. Same for the next line.

src/Dialog.cc
```diff
@@ -75,3 +80,141 @@ QQuickItem *Dialog::RootItem() const
...
+  std::string value {""};
+  std::string config = "<?xml version=\"1.0\"?>\n\n";
+  tinyxml2::XMLPrinter defaultPrinter;
+  bool configExists{true};
+  std::string dialogName = this->objectName().toStdString();
+
+  auto Value = [&_attribute, &doc, &dialogName]()
+  {
+    // Process each dialog
+    // If multiple attributes share the same name, return the last one
```
chapulina, line 151: It looks like the new logic actually returns the first one, right?

src/Dialog_TEST.cc
chapulina: Thanks for adding it for the first test case. Can you add those checks for all the others? It's very useful to be explicit about the exact case that's being tested.

mayman99: Make sense :+1:

src/Dialog.cc
mayman99: copy pasting comments without re-reading :upside_down_face: 
thanks for all the catches!

mayman99:
@chapulina green :+1: 
I didn't have time to convert the Read/Update functions to use template functions (I did but it results in a seg fault when used in `gz-sim`, my maintenance time was not enough to find the problem)

chapulina (approved):

PR #419, opened 2022-06-16
Improve workflow tests robustness

Blast545:
Attempts to fix #58, WIP

## Summary
Before trying to change the underlying virtual buffer used to run GUI tests, this PR changes the arguments passed to the after_make.sh script to make it more robust.

This is already done in gz-rendering, see: https://github.com/gazebosim/gz-rendering/pull/255

Running CI a couple of times before asking for reviews.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests -> potentially fixes tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

PR #422, opened 2022-06-16
[ign -> gz] CMake functions

chapulina:
* Part of https://github.com/gazebo-tooling/release-tools/issues/719

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package(ignition-cmake3 REQUIRED)
 #============================================================================
 # Configure the project
 #============================================================================
-ign_configure_project(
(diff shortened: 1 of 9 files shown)


methylDragon (approved):
Stack builds with this :ok_hand:

methylDragon:
I'm not sure if I saw: `39 - INTEGRATION_Examples_TEST (Subprocess aborted)` before

chapulina:
> I'm not sure if I saw: 39 - INTEGRATION_Examples_TEST (Subprocess aborted) before

That's flaky, it may be better once we forward-port #419 🤞🏽 One way to verify this is that it failed in one of the Jammy builds but not the other.

PR #449, opened 2022-08-01
[ign to gz] Deduplicate QML files

chapulina:
# 🦟 Bug fix

* Follow up to https://github.com/gazebosim/gz-gui/pull/395
* Part of https://github.com/gazebo-tooling/release-tools/issues/711

## Summary

Here's a migration for the QML files which removes the duplicated ones. The trick is:

* `include/gz/gui/qml/qmldir_deprecated` adds support for the `ignition.gui` deprecated module, so downstream QML files calling `import ignition.gui 1.0` still work
* The aliases added inside `qmldir` and `qmldir_deprecated` make sure that both `Ign` and `Gz` components are available even when importing either `ignition.gui` or `gz.gui`.

## Test it

* Build and run the `gz_components` example - this shows that both `IgnSpinBox` and `GzSpinBox` work with `import gz.gui 1.0`
* Then change it to `import ignition.gui 1.0` and check that this works too

## Checklist
- [ ] Added tests


methylDragon (approved):
LGTM on green CI, I'm not sure why CI isn't green though...

Also noting that this means the other Ign prefixed stuff won't work in downstream libs since the redirection aliases for them :boom: ?

(The Ign -> Gz aliases were in https://github.com/gazebosim/gz-gui/pull/449/files#diff-978089d7a839a8b0ec51044f445624a06765881a90adaa0e66a381d1326bcc3b , but now are gone)

chapulina:
> I'm not sure why CI isn't green though...

* That's caused by https://github.com/gazebosim/gz-common/pull/393. I haven't had time to look into it yet

> the other Ign prefixed stuff won't work in downstream libs since the redirection aliases for them

Ah yeah, the ones outside of the module aren't meant to be used downstream, so that should be ok

PR #455, opened 2022-08-05
Fix attribute update when <dialog> isn't present

chapulina:
# 🦟 Bug fix

## Summary

This fixes a situation when a file doesn't have a `<dialog>` element yet, and we need to create one to update its value.

I also increased the `Dialog`'s test coverage to 100% and pre-deprecated a function that's no longer needed.

## Checklist
- [x] Added tests
- [x] Consider updating Python bindings (if the library has them)

### DIFF
--- a/src/Dialog.cc
+++ b/src/Dialog.cc
@@ -27,9 +27,6 @@ namespace ignition
   {
     class DialogPrivate
     {
-      /// \brief default dialog config
-      public: std::string config{""};
-
       /// \brief Pointer to quick window
       public: QQuickWindow *quickWindow{nullptr};
     };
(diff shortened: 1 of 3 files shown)


src/Dialog.cc
```diff
@@ -101,37 +103,43 @@ bool Dialog::UpdateConfigAttribute(const std::string &_path,
...
 
   return true;
 }
 
 /////////////////////////////////////////////////
-void Dialog::SetDefaultConfig(const std::string &_config)
+void Dialog::SetDefaultConfig(const std::string &)
 {
-  this->dataPtr->config = _config;
+  ignwarn << "Dialog::SetDefaultConfig has no effect." << std::endl;
```
mayman99, line 142: This means the only way to write to the dialog config is to use the `Update` function, right?

mayman99: I imagine if the config has many fields it might be not the best way, or is that not a major use case?

chapulina: Yeah I believe that's the only way right now. If the user wants to set multiple attributes, they'd need to set them one by one. And there's no interface for adding elements, etc.

mayman99 (approved):
lgtm!
the only two downsides to the new impl is that we can only write `bool` attributes and we only write one by one, we can update it if such use cases arises then imho

PR #459, opened 2022-08-08
Fix main window moving with ColorDialogue of GzColor

AzulRadio:
Signed-off-by: youhy <haoyuan2019@outlook.com>

# Bug fix

## Summary

Bug: dragging the color dialogue will cause the main window to move with the dialogue window, which was not the behavior before the common widget GzColor was added.

gzcolor_bug_demo

With this fix:

gzcolor_bugfix_demo

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gui/qml/GzColor.qml
+++ b/include/ignition/gui/qml/GzColor.qml
@@ -57,7 +57,6 @@ Item {
     title: "Choose a color"
     visible: false
     showAlphaChannel: true
-    modality: Qt.ApplicationModal
     onAccepted: {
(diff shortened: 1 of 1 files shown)


chapulina (approved):
Nice catch, it's useful for this dialog not to be modal so that the user can interact with the 3D scene while choosing a color.

We could consider changing the `onAccepted` slot to `onCurrentColorChanged`, so that the color is changed when the user chooses different colors with the dialog open, not only when they close the dialog. That has performance implications though, so I think it's ok to keep it as it is.

PR #357, opened 2022-02-19
Add fallback rendering for other APIs

darksylinc:
# 🦟 Bug fix

No ticket has been assigned for this issue.

## Order in which PRs must be merged

1. gazebosim/gz-rendering#706
2. This PR.
    - or optionally, merge this one as part of the next one aka PR 467, which includes it
3. gazebosim/gz-gui#467
4. gazebosim/gz-sim#1735

## Summary

Until we get native API to Qt implemented, fallback system can be used
to transfer GPU (Ogre) -> CPU -> GPU (Qt)

This is important for getting APIs other than OpenGL to work with Qt.

This is slow, but it works.

Signed-off-by: Matias N. Goldberg <dark_sylinc@yahoo.com.ar>

## Dependency

After the CI failed, I realized this PR depends on either ignitionrobotics/ign-rendering#553 or ignitionrobotics/ign-rendering#565 and thus cannot be merged until then.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

Issue #17, opened 2017-12-26
Properly terminate render engine


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gui/issues/17)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
@iche033 mentioned on [pull request #62](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gui/pull-requests/62) that we should call `Fini` on the ignition-rendering engine before terminating the program so it shuts down properly.

I tried calling fini when the engine has no more scenes like this:

~~~
diff -r 17c6c8bc83ab46edfadd765870d150e87d868e4e src/plugins/Scene3D.cc
--- a/src/plugins/Scene3D.cc	Tue Dec 26 08:53:22 2017 -0800
+++ b/src/plugins/Scene3D.cc	Tue Dec 26 09:27:22 2017 -0800
@@ -90,6 +90,8 @@
     engine->DestroyScene(scene);
 
     // TODO: If that was the last scene, terminate engine?
+    if (engine->SceneCount() == 0)
+      engine->Fini();
   }
 }
~~~

But if after doing this, we try to open a new scene plugin (i.e. re-init the engine), the program crashes with this backtrace:

~~~
/usr/local/lib/libignition-rendering-ogre.so.0(_ZN8ignition9rendering9OgreScene13CreateContextEv+0x75) [0x7f49e1e5c731] /home/louise/code/ign-rendering/src/ogre/OgreScene.cc:339
/usr/local/lib/libignition-rendering-ogre.so.0(_ZN8ignition9rendering9OgreScene8InitImplEv+0x3a) [0x7f49e1e57faa] /home/louise/code/ign-rendering/src/ogre/OgreScene.cc:124
/usr/local/lib/libignition-rendering-base.so.0(_ZN8ignition9rendering9BaseScene4InitEv+0x1d1) [0x7f49e18de6a7] /home/louise/code/ign-rendering/src/base/BaseScene.cc:70
[...]

PR #467, opened 2022-08-14
Add Vulkan QML backend

darksylinc:
# 🎉 New feature

## Order in which PRs must be merged

1. gazebosim/gz-rendering#706
2. gazebosim/gz-gui#357
    - or optionally, merge this one as part of the next one aka PR 467, which includes it
3. This PR.
4. gazebosim/gz-sim#1735

## Summary

> **_Note:_** This PR depends on gazebosim/gz-rendering#706

> **_Note:_** This PR is polluted with changes that are supposed to be merged by PR #357 

This PR adds native Vulkan backend using Qt's QML Vulkan RHI.

I couldn't believe it so I opened it up on RenderDoc and indeed it's using Vulkan:

Screenshot_2022-08-14_19-51-44

This PR breaks ABI slightly, so it's best if it targets post-Garden.

With a few global variables we could workaround it and backport it to Garden though. Note that Garden supports Vulkan via a fallback that is very slow. This PR offers native, full speed integration.

> **_Note:_** Qt 5.14 is the minimum requirement, which means it's not available on Ubuntu 20.04 but rather Ubuntu 22.04. I put macros to use at least 5.15.2 because that's what I tested against.

> **_Note:_** Qt 5 considers Vulkan QML interface to be 'experimental' and I've submitted a [patch to Qt](https://bugreports.qt.io/browse/QTBUG-105158) to improve compliance with the Vulkan standard. Ideally Gazebo should eventually migrate to Qt 6.

## Missing tasks:

- [x] Ensure code compiles without Qt 5.15.2+
- [x] Ensure code compiles when Vulkan isn't available
- [x] Gazebo needs to tell OgreNext to make a layout transition / synchronization for Qt (via BarrierSolver).
- [x] Run Vulkan validations to catch any potential issue
    - Done. Only validation errors show up on shutdown, but that's difficult to debug right now because gz-gui crashes on shutdown due to infamous crash at shutdown (can't find the ticket now) that could be the cause. May also be a side effect of #17 
- [ ] Add vulkan toggle via command line

## Test it

This PR can be tested once gazebosim/gz-sim#1735 (which comes after this PR) is merged, which adds the capability to enable the features in this PR via CLI.

### Testing it in Ubuntu 20.04

It is possible to build with Ubuntu 20.04 against Qt 5.15.2. You need the following:

1. Install Qt 5.15.2 using the [online installer](https://download.qt.io/official_releases/online_installers/)
2. [Compile Gazebo](https://gazebosim.org/docs/garden/install_ubuntu_src) with: `colcon build --cmake-args -DCMAKE_PREFIX_PATH=/opt/Qt/5.15.2/gcc_64 -DQT_QMAKE_EXECUTABLE=/opt/Qt/5.15.2/gcc_64/bin/qmake --merge-install`
3. To run gazebo go to the install folder and do:
```bash
export LD_LIBRARY_PATH=/opt/Qt/5.15.2/gcc_64/lib
. setup.bash
gz sim shapes.sdf
```

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -22,6 +22,7 @@
 #include "MinimalSceneRhi.hh"
 #include "MinimalSceneRhiMetal.hh"
 #include "MinimalSceneRhiOpenGL.hh"
+#include "MinimalSceneRhiVulkan.hh"
 
 #include <algorithm>
 #include <list>
@@ -39,6 +40,7 @@
 #include <gz/rendering/Camera.hh>
 #include <gz/rendering/RayQuery.hh>
 #include <gz/rendering/RenderEngine.hh>
+#include <gz/rendering/RenderEngineVulkanExternalDeviceStructs.hh>
 #include <gz/rendering/RenderingIface.hh>
 #include <gz/rendering/Scene.hh>
 #include <gz/rendering/Utils.hh>
@@ -49,6 +51,10 @@
 #include "gz/gui/GuiEvents.hh"
 #include "gz/gui/Helpers.hh"
 #include "gz/gui/MainWindow.hh"
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 2)
+#  include <QVulkanInstance>
+#endif
 
 Q_DECLARE_METATYPE(gz::gui::plugins::RenderSync*)
 
@@ -298,8 +304,19 @@ void RenderSync::Shutdown()
 GzRenderer::GzRenderer()
   : dataPtr(utils::MakeUniqueImpl<Implementation>())
 {
-  // Set default graphics API to OpenGL
-  this->SetGraphicsAPI(rendering::GraphicsAPI::OPENGL);
+  const std::string backendApiName = gz::gui::renderEngineBackendApiName();
+  if (backendApiName == "vulkan")
+  {
+    this->SetGraphicsAPI(rendering::GraphicsAPI::VULKAN);
+  }
+  else if (backendApiName == "metal")
+  {
+    this->SetGraphicsAPI(rendering::GraphicsAPI::METAL);
+  }
+  else
+  {
+    this->SetGraphicsAPI(rendering::GraphicsAPI::OPENGL);
+  }
 }
 
 /////////////////////////////////////////////////
@@ -562,6 +579,87 @@ void GzRenderer::BroadcastKeyPress()
 }
 
 /////////////////////////////////////////////////
+rendering::CameraPtr GzRenderer::Camera()
+{
+  return this->dataPtr->camera;
+}
+
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 2)
+/////////////////////////////////////////////////
+/// \brief fillQtInstanceExtensionsToOgre
+/// Extract Vulkan Instance extension information to be sent to OgreNext
+/// \param[in] inst Qt's Vulkan Instance to extract
+/// \param[out] externalInstance Data to be sent to OgreNext
+static void fillQtInstanceExtensionsToOgre(
+  const QVulkanInstance *inst,
+  rendering::GzVulkanExternalInstance &externalInstance)
+{
+  {
+    QByteArrayList extensions = inst->extensions();
+
+    for (const auto &ext : extensions)
+    {
+      externalInstance.instanceExtensions.push_back(VkExtensionProperties{});
+      VkExtensionProperties &extProp =
+        externalInstance.instanceExtensions.back();
+      strncpy(extProp.extensionName, ext.constData(),
+              VK_MAX_EXTENSION_NAME_SIZE);
+      extProp.extensionName[VK_MAX_EXTENSION_NAME_SIZE - 1u] = 0;
+    }
+  }
+
+  {
+    QByteArrayList layers = inst->layers();
+
+    for (const auto &layer : layers)
+    {
+      externalInstance.instanceLayers.push_back(VkLayerProperties{});
+      VkLayerProperties &layerProp = externalInstance.instanceLayers.back();
+      strncpy(layerProp.layerName, layer.constData(),
+              VK_MAX_EXTENSION_NAME_SIZE);
+      layerProp.layerName[VK_MAX_EXTENSION_NAME_SIZE - 1u] = 0;
+    }
+  }
+}
+
+/////////////////////////////////////////////////
+/// \brief fillQtDeviceExtensionsToOgre
+/// Extract Vulkan Device extension info to be sent to OgreNext
+/// This data is obtained from Environment variables
+/// \param[out] externalDevice Data to be sent to OgreNext
+static void fillQtDeviceExtensionsToOgre(
+  rendering::GzVulkanExternalDevice &externalDevice)
+{
+  // We know Qt adds these by looking at
+  //  qt-everywhere-src-5.15.2/qtbase/src/gui/rhi/qrhivulkan.cpp
+  QVector<QString> deviceExtensions;
+  deviceExtensions.append( "VK_KHR_swapchain" );
+
+  QByteArray envExts = qgetenv( "QT_VULKAN_DEVICE_EXTENSIONS" );
+  if( !envExts.isEmpty() )
+  {
+    QByteArrayList envExtList = envExts.split( ';' );
+    for( const auto &ext : envExtList )
+    {
+      if( !ext.isEmpty() )
+      {
+        deviceExtensions.append( ext );
+      }
+    }
+  }
+
+  for( const auto &ext : deviceExtensions )
+  {
+    externalDevice.deviceExtensions.push_back( VkExtensionProperties{} );
+    VkExtensionProperties &extProp = externalDevice.deviceExtensions.back();
+    strncpy(extProp.extensionName, ext.toStdString().c_str(),
+            VK_MAX_EXTENSION_NAME_SIZE);
+    extProp.extensionName[VK_MAX_EXTENSION_NAME_SIZE - 1u] = 0;
+  }
+}
+#endif
+
+/////////////////////////////////////////////////
 std::string GzRenderer::Initialize(RenderThreadRhi &_rhi)
 {
   if (this->initialized)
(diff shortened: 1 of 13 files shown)


src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -537,7 +555,88 @@ void GzRenderer::BroadcastKeyPress()
...
+static void fillQtDeviceExtensionsToOgre(
+  rendering::GzVulkanExternalDevice &externalDevice)
+{
+  // We know Qt adds these by looking at
+  //  qt-everywhere-src-5.15.2/qtbase/src/gui/rhi/qrhivulkan.cpp
+  QVector<QString> deviceExtensions;
+  deviceExtensions.append( "VK_KHR_swapchain" );
+
+  QByteArray envExts = qgetenv( "QT_VULKAN_DEVICE_EXTENSIONS" );
+  if( !envExts.isEmpty() )
```
ahcorde, line 615:
```suggestion
  if(!envExts.isEmpty())
```

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -537,7 +555,88 @@ void GzRenderer::BroadcastKeyPress()
...
+  {
+    QByteArrayList envExtList = envExts.split( ';' );
+    for( const auto &ext : envExtList )
+    {
+      if( !ext.isEmpty() )
```
ahcorde, line 620:
```suggestion
    for(const auto &ext : envExtList)
    {
      if(!ext.isEmpty())
```

src/plugins/minimal_scene/MinimalSceneRhiVulkan.cc
```diff
@@ -0,0 +1,247 @@
...
+
+/////////////////////////////////////////////////
+void GzCameraTextureRhiVulkan::Update(rendering::CameraPtr _camera)
+{
+  // It says Metal but it also works for Vulkan in the exact same way
+  _camera->RenderTextureMetalId(&this->dataPtr->textureId);
+}
+
+/////////////////////////////////////////////////
+/////////////////////////////////////////////////
```
ahcorde, line 90:
```suggestion
```

iche033:
> This PR breaks ABI slightly, so it's best if it targets post-Garden.

I think there is no ABI breakage? Adding a new function to `Helpers.hh` should be ok. The minimal scene plugin headers are not installed.

src/plugins/minimal_scene/MinimalSceneRhiVulkan.cc
darksylinc: The double comment is on purpose (it's also in MinimalSceneRhiOpenGL.cc.cpp & MinimalSceneRhiMetal.mm) to distinguish the switch from `GzCameraTextureRhiVulkan` class definitions to `RenderThreadRhiVulkan`.

(fun fact: GzCameraTextureRhi* don't seem to actually be used. It seems to be dead code but this theory needs to be tested; I added a TODO in `class GzCameraTextureRhi`).

iche033:
we need to bump gz-rendering dep version to 8.0.0 and update CI configurations. I'll do that in a separate PRs after all changes are forward ported to `main`.

mjcarroll:
> we need to bump gz-rendering dep version to 8.0.0 and update CI configurations. I'll do that in a separate PRs after all changes are forward ported to main.

I thought this could land in garden?

iche033:
> I thought this could land in garden?

https://github.com/gazebosim/gz-rendering/pull/706 may have some ABI breaking changes. We could try to work around it.

darksylinc:
Originally yes. With a few hacks I think/thought it could still land on Garden.

However my original assessment was assuming #357 was already merged (I was surprised when I found out it was not), and AFAIK Garden will be released without it.

I'm not sure if we can avoid ABI breakage given that #357 wasn't included. I'd have to carefully look into it if we want to backport at least the fallback path.

iche033:
I think https://github.com/gazebosim/gz-gui/pull/357 can go in gz-gui6 and it touches only files in the minimal scene GUI plugin. Those headers are not installed so we don't have to worry about API/ABI compatibility.

iche033:
coming back to this, I think https://github.com/gazebosim/gz-rendering/pull/706 is going to be hard to backport to garden. @mjcarroll are you ok with these PRs landing in `main`?

mjcarroll:
Yes, lets land them on main for expedience.

azeey:
needs version bumps

darksylinc:
@azeey I'm trying to rebase this PR to main; but current main has a gazillion build errors.

`gz-gui/main` branch clearly needs to merge with `gz-gui/gz-gui7` branch.

gz-sim/main is in the same state (merging has quite a number of merge conflicts).
gz-launch/main is in the same state.

On top of those build errors, they're also looking for the wrong project versions (e.g they look for gz-rendering7 instead of gz-rendering8).

Basically building `Harmonic` via [collection-harmonic.yaml](https://github.com/gazebo-tooling/gazebodistro/blob/master/collection-harmonic.yaml) doesn't yet work out of the box.

Should I wait until Harmonic is buildable?

azeey:
> Should I wait until Harmonic is buildable?
Yes, there are a couple of PRs in flight bumping the major versions of `gz-rendering` and its dependees. These are tracked in https://github.com/gazebo-tooling/release-tools/issues/853. Let's wait till all of those are merged.

iche033:
after https://github.com/gazebosim/gz-gui/pull/512 gets in, everything up to gz-gui should be buildable

darksylinc:
Another rebased and retargeted.

darksylinc:
I need to rebase this PR now that the previous one has been merged.

darksylinc:
Are the tests flaky? Jammy always fails on these two:

>   	 51 - INTEGRATION_minimal_scene (Timeout)
>  	 53 - INTEGRATION_transport_scene_manager (SEGFAULT)

And Focal always succeeded except this last time (I made no significant changes that would cause this, possibly flaky):

>  	 45 - INTEGRATION_camera_tracking (SEGFAULT)

As for Windows CI, I fixed the problems caused by my PR; the new warnings don't look like caused by me (?)

And macOS was failing build, I fixed that. Now it fails various tests. I don't know if those are my fault.

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -561,6 +578,87 @@ void GzRenderer::BroadcastKeyPress()
...
+      }
+    }
+  }
+
+  for (const auto &ext : deviceExtensions)
+  {
+    externalDevice.deviceExtensions.push_back(VkExtensionProperties{});
+    VkExtensionProperties &extProp = externalDevice.deviceExtensions.back();
+    strncpy(extProp.extensionName, ext.toStdString().c_str(),
+            VK_MAX_EXTENSION_NAME_SIZE);
```
iche033, line 656: there is a compile warning here:

```
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘void fillQtDeviceExtensionsToOgre(gz::rendering::v8::GzVulkanExternalDevice&)’ at /home/ian/code/gz_h_ws/src/gz-gui/src/plugins/minimal_scene/MinimalScene.cc:655:12,
    inlined from ‘std::string gz::gui::plugins::GzRenderer::Initialize(gz::gui::plugins::RenderThreadRhi&)’ at /home/ian/code/gz_h_ws/src/gz-gui/src/plugins/minimal_scene/MinimalScene.cc:711:35:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: warning: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ specified bound 256 equals destination size [-Wstringop-truncation]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));

```sh

**Review by iche033 — COMMENTED** [review]
tested this with https://github.com/gazebosim/gz-sim/pull/1735 on Jammy with Qt 5.13.3. It's working well for me with worlds like `sensors_demo.sdf` and `segmetnation_camera.sdf`. 

One thing I noticed is that gui crashes when I tried to resize the window. If it's not trivial to fix, we can ticket an issue for this.

**Comment by iche033** [comment]
> 51 - INTEGRATION_minimal_scene (Timeout)
> 53 - INTEGRATION_transport_scene_manager (SEGFAULT)
> 45 - INTEGRATION_camera_tracking (SEGFAULT)

I've seen these tests fail from time to time. I think they are flaky. 

> And macOS was failing build, I fixed that. Now it fails various tests. I don't know if those are my fault.

I haven't seen those tests failed before. Not sure why. I restarted a new homebrew build and now it only has one expected test failure: https://build.osrfoundation.org/job/ignition_gui-ci-pr_any-homebrew-amd64/1893/. Will need to keep an eye on this.

File: src/plugins/minimal_scene/MinimalScene.cc (hunk 4 above)
Comment (darksylinc) [reply src/plugins/minimal_scene/MinimalScene.cc:656]: That should be impossible? The macros:

```cpp
#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 2) && QT_CONFIG(vulkan)
#endif
```

Which means it should be excluded from complication with Qt 5.13.3

Anyway, the macro says we are doing this:

```cpp
char data[256];
strncpy( data, somePtr, 256);
```

Which is dangerous because if strlen( somePtr ) >= 256; then data[255] will not contain the null terminator `\0`.

Which is why I do on the next line:

```cpp
extProp.extensionName[VK_MAX_EXTENSION_NAME_SIZE - 1u] = 0;
```

in order to enforce the string is always valid.

Anyway, there may be other ways to fix this warning:

```cpp
strncpy(extProp.extensionName, ext.constData(),
              VK_MAX_EXTENSION_NAME_SIZE - 1u);
```

iche033: ah typo, I was testing with Qt 5.1**5**.3 not Qt 5.1**3**.3. I added the suggested fix to that line (and left other similar lines untouched) to get rid of the warning.  99f887f

iche033 (approved):
CI looks good. Tested together with https://github.com/gazebosim/gz-sim/pull/1735:

```
gz sim -v 4 --render-engine-gui-api-backend vulkan  sensors_demo.sdf
```
GUI works for me in both vulkan and OpenGL.

PR #476, opened 2022-08-24
Added Reset button to world_control

ahcorde:

# 🎉 New feature

## Summary

Added a reset button to the world Control plugin
reset_button

## Test

```bash
gz sim rolling_shapes.sdf
```

## Checklist
- [ ] Added tests


mjcarroll:
A super pedantic nit:  Can we get a reset icon? :)

src/plugins/world_control/WorldControl.cc
```diff
@@ -306,6 +306,18 @@ void WorldControl::OnPause()
   this->dataPtr->SendEventMsg(msg);
 }
 
+/////////////////////////////////////////////////
+void WorldControl::OnReset()
+{
+  gz::msgs::WorldControl msg;
+  auto msgReset = new gz::msgs::WorldReset();
```
jennuine, line 313: nit: already in the `gz` namespace
```suggestion
  msgs::WorldControl msg;
  auto msgReset = new msgs::WorldReset();
```

src/plugins/world_control/WorldControl.qml
```diff
@@ -82,6 +92,24 @@ RowLayout {
    */
   property bool paused: false
 
+  /**
+   * Reset
+   */
+  RoundButton {
+    id: stopButton
+    objectName: "stopButton"
+    visible: showReset
```
jennuine, line 101: Should reset only be visible if the sim is paused?

src/plugins/world_control/WorldControl.qml
```diff
@@ -82,6 +92,24 @@ RowLayout {
    */
   property bool paused: false
 
+  /**
+   * Reset
+   */
+  RoundButton {
```
jennuine, line 98: Can we add a hovered tooltip that states what reset does?

jennuine (commented):
> Can we get a reset icon?
+1, the square is misleading. At a glace, it looks like a stop button. 

Also, since we are feature frozen, this will need to come in a minor release

ahcorde:
> A super pedantic nit: Can we get a reset icon? :)

@mjcarroll I was thinking about adding an icon but I was trying to follow the same style which is an unicode geometric shape https://en.wikipedia.org/wiki/List_of_Unicode_characters. I'm open to suggestion.

Anyhow the way that I implemented the `reset` is more like a `stop` (reset + pause)

jennuine:
How about using the same icon as refresh in `ImageDisplay`? https://github.com/gazebosim/gz-gui/blob/a60cb7d61339b407ff3285ddbf16f618c4c4b3b9/src/plugins/image_display/ImageDisplay.qml#L63

ahcorde:
> How about using the same icon as refresh in `ImageDisplay`?
> 
> https://github.com/gazebosim/gz-gui/blob/a60cb7d61339b407ff3285ddbf16f618c4c4b3b9/src/plugins/image_display/ImageDisplay.qml#L63

nice! thank you @jennuine

jennuine:
Users may accidentally click this icon so another nice feature would be to have a confirmation dialog before resetting similar to https://github.com/gazebosim/gz-gui/pull/225 (with the ability to disable the dialog through the config)

chapulina (commented):
It would be nice to also add the `Ctrl+R` shortcut, and then we can close

* https://github.com/gazebosim/gz-sim/issues/203

scpeters:
@osrf-jenkins run tests please

ahcorde:
> It would be nice to also add the `Ctrl+R` shortcut, and then we can close
> 
> * [gazebosim/gz-sim#203](https://github.com/gazebosim/gz-sim/issues/203)

https://github.com/gazebosim/gz-gui/pull/476/commits/dbd8c74f31839dd4186bc074e2d0c30ed4748f1c

src/plugins/world_control/WorldControl.cc
ahcorde: https://github.com/gazebosim/gz-gui/pull/476/commits/dbd8c74f31839dd4186bc074e2d0c30ed4748f1c

src/plugins/world_control/WorldControl.qml
ahcorde: https://github.com/gazebosim/gz-gui/pull/476/commits/639643dc602fa71686806d2188548c14f254461b

src/plugins/world_control/WorldControl.qml
ahcorde: you might want to reset when the simulation is running

ahcorde:
friendly ping @jennuine @mjcarroll

jennuine (commented):
nit: can we leave the play/step buttons how they were and add the reset button to the right of the step button? And have the reset button be the same size as step?

ahcorde:
@jennuine 

what do you think about this ?

Selection_094

mjcarroll (approved):

jennuine:
> @jennuine
> 
> what do you think about this ?
> 
> Selection_094

Looks great, thanks! How do you and @mjcarroll feel about the reset button being to the right of the step button? Reasons being:
1. Users are less likely to accidentally hit reset when attempting to pause
2. It keeps the old/previous layout the same while adding the new feature

ahcorde:
> > @jennuine
> > what do you think about this ?
> > Selection_094
> 
> Looks great, thanks! How do you and @mjcarroll feel about the reset button being to the right of the step button? Reasons being:
> 
> 1. Users are less likely to accidentally hit reset when attempting to pause (I know this is less likely but I can still see it happening)
> 2. It keeps the old/previous layout the same while adding the new feature

Anyhow when you click in the `reset` button a Dialog will appers to ensure that you want to reset the simulation. Not dangerous at all.

ahcorde:
> > > @jennuine
> > > what do you think about this ?
> > > Selection_094
> > 
> > 
> > Looks great, thanks! How do you and @mjcarroll feel about the reset button being to the right of the step button? Reasons being:
> > 
> > 1. Users are less likely to accidentally hit reset when attempting to pause (I know this is less likely but I can still see it happening)
> > 2. It keeps the old/previous layout the same while adding the new feature
> 
> Anyhow when you click in the `reset` button a Dialog will appers to ensure that you want to reset the simulation. Not dangerous at all.

changing layout doesn't seem to be working, the step botton uses a mouseclick area, it's tricky to change it

mjcarroll:
> the step botton uses a mouseclick area, it's tricky to change it

If it is too difficult to fix, I think that the dialog confirmation is sufficiently safe.

ahcorde:

ahcorde:

ahcorde:

PR #479, opened 2022-08-26
Clean up library.hh headers

chapulina:
# 🦟 Bug fix

* Needed by https://github.com/gazebosim/gz-transport/pull/315

## Summary

* Include strictly necessary headers instead of `library.hh`.
* Remove Windows warning suppressions that aren't needed anymore.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/gui/Conversions.hh
+++ b/include/gz/gui/Conversions.hh
@@ -18,13 +18,8 @@
 #ifndef GZ_GUI_CONVERSIONS_HH_
 #define GZ_GUI_CONVERSIONS_HH_
 
-#ifdef _MSC_VER
-#pragma warning(push, 0)
-#endif
 #include <gz/msgs/time.pb.h>
-#ifdef _MSC_VER
-#pragma warning(pop)
-#endif
+
 #include <gz/common/KeyEvent.hh>
 #include <gz/common/MouseEvent.hh>
 #include <gz/math/Color.hh>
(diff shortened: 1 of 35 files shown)


chapulina:
It looks like there are some Windows warnings still leaking through `moc_TopicEcho.cpp`, even though they should be suppressed by

https://github.com/gazebosim/gz-gui/blob/de0c56be10c98ca6497920941cd81f5fab0697a1/src/plugins/topic_echo/TopicEcho.hh#L21-L27

🤔

mjcarroll:
I ran into this somewhere else, I don't think that `#pragma warning(push, 0)`  is valid...

From the [docs](https://docs.microsoft.com/en-us/cpp/preprocessor/warning?view=msvc-170)

> The warning pragma also supports the following syntax, where the optional n parameter represents a warning level (1 through 4).

jennuine:
> It looks like there are some Windows warnings still leaking through moc_TopicEcho.cpp, even though they should be suppressed

Would they be fixed if we added:
```c++
#pragma warning(disable: 4251)
```
?

mjcarroll:
> Would they be fixed if we added:

Yes, this is the correct mechanism, it's what I had to do elsewhere.

chapulina:
> Would they be fixed if we added:

Done in [d7aaecf](https://github.com/gazebosim/gz-gui/pull/479/commits/d7aaecfa3111e73d4e00a6a9230cea0e51911036), hopefully Windows CI will be green now 🤞🏽

mjcarroll (approved):

jennuine (approved):

PR #490, opened 2022-09-14
Remove unused headers

jennuine:
# 🦟 Bug fix

## Summary

Remove unused headers, which were causing the abi checker to fail (first detected in https://github.com/gazebosim/gz-gui/pull/486)

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #491, opened 2022-09-14
JUST FOR TESTING Remove redundant auto header and its redirection header

methylDragon:
From https://github.com/gazebosim/gz-physics/pull/305 we know that `gz_auto_header.hh.in` files are not being used. The problem is that they're not being used AND there is a redirection header that points to nowhere because of it.

This causes the ABI checker to fail (surfaced by https://github.com/gazebosim/gz-gui/pull/486). So this PR fixes that issue by removing both files.


methylDragon:
Closing because it's a duplicate of https://github.com/gazebosim/gz-gui/pull/490 (oops)

PR #494, opened 2022-10-06
Tries to fix a scenario when a user sends a malformed point cloud

arjo129:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
This PR fixes a rather odd bug. Essentially if a user were to malform a pointcloud by claiming the point cloud to contain only "x,y,z" fields but has a datasize larger than that, the condition `iterX == iterX.End() && iterY == iterY.End() && iterZ == iterZ.End();` does not get met. Thus the system will keep iterating till a segfault comes along. This PR enforces better check to make sure we do not segfault.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/point_cloud/PointCloud.cc
+++ b/src/plugins/point_cloud/PointCloud.cc
@@ -406,8 +406,7 @@ void PointCloudPrivate::PublishMarkers()
   {
     // Value from float vector, if available. Otherwise publish all data as
     // zeroes.
-    float dataVal = 0.0;
-    dataVal = this->floatVMsg.data(ptIdx);
+    float dataVal = this->floatVMsg.data(ptIdx);
 
     // Don't visualize NaN
     if (std::isnan(dataVal))
--- a/src/plugins/point_cloud/PointCloud.cc
+++ b/src/plugins/point_cloud/PointCloud.cc
@@ -17,6 +17,7 @@
 
 #include "gz/msgs/pointcloud_packed.pb.h"
 
+#include <algorithm>
 #include <limits>
 #include <string>
 #include <utility>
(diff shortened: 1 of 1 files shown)


src/plugins/point_cloud/PointCloud.cc
```diff
@@ -388,17 +389,24 @@ void PointCloudPrivate::PublishMarkers()
...
+
+  for (; ptIdx < std::min<int>(this->floatVMsg.data().size(), num_points);
+    ++iterX, ++iterY, ++iterZ, ++ptIdx)
   {
     // Value from float vector, if available. Otherwise publish all data as
     // zeroes.
     float dataVal = 0.0;
-    if (this->floatVMsg.data().size() > ptIdx)
-    {
-      dataVal = this->floatVMsg.data(ptIdx);
```
hidmic, line 400: @arjo129 I understand why the `if` clause is no longer necessary, but why remove the assignment too?

hidmic (changes requested):
`codecov` wants tests :)

src/plugins/point_cloud/PointCloud.cc
arjo129: Oh no that was not intended :skull: will add some unit tests as well.

arjo129: My bad. This is wrong :skull: fixed.

hidmic (approved):

azeey:
@arjo129 The PointCloud plugin currently does not work if a corresponding float_v message is not available. Reverting this PR fixes that on the gz-gui10 branch. Mind taking a look?

PR #502, opened 2022-11-11
Fix large / unexpected camera movements

iche033:
Signed-off-by: Ian Chen <ichen@osrfoundation.org>

# 🦟 Bug fix

Related issue: https://github.com/gazebosim/gz-sim/issues/1085

## Summary
Fixes large / unexpected camera movement when orbiting / panning due to missing mouse events.

More info:

The `Minimal Scene` is responsible for propagating all mouse events to other GUI plugins. However, mouse events were continuously overriden by the latest ones before they can be propagated / processed. This causes important events like a mouse press to be lost. The `Interactive View Control` relies on mouse press pos to determine the anchor point for camera orbiting and panning, so if it misses a mouse press pos, it justs uses the last anchor point for camera movements. When the anchor point is far away, it causes large camera jumps. 

This gif shows the problem. The first pan motion successfully sets the anchor point (yellow ellipsoid). Subsequent camera orbit and pan motions miss the mouse press events and thus they still use the old anchor point that's on top of the red box.

missing_mouse_events

This PRs fixes the issue by storing mouse events in a list in `Minimal Scene` instead of just the latest event, and makes sure that the `Interactive View Control` always handles mouse press events.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -77,7 +77,12 @@ class ignition::gui::plugins::IgnRenderer::Implementation
   /// \brief Key event
   public: common::KeyEvent keyEvent;
 
-  /// \brief Max number of mouse events to process
+  /// \brief Max number of mouse events to store in the queue.
+  /// These events are then propagated to other gui plugins. A queue is used
+  /// instead of just keeping the latest mouse event so that we can capture
+  /// important events like mouse presses. However we keep the queue size
+  /// small on purpose so that we do not flood other gui plugins with events
+  /// that may be outdated.
   public: const unsigned int kMaxMouseEventSize = 5u;
 
   /// \brief Mutex to protect mouse events
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -67,11 +67,17 @@ class ignition::gui::plugins::IgnRenderer::Implementation
   /// \brief Flag to indicate if drop event is dirty
   public: bool dropDirty{false};
 
-  /// \brief Mouse event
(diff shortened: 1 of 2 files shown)


ahcorde (changes requested):
linters

iche033:
> linters

fixed. [23d8993](https://github.com/gazebosim/gz-gui/pull/502/commits/23d899341fbbfcb8c435eb87bb59aeb378cdad3b)

ahcorde (approved):

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -67,12 +68,18 @@ class ignition::gui::plugins::IgnRenderer::Implementation
...
   public: common::MouseEvent mouseEvent;
 
+  /// \brief A list of mouse events
+  public: std::list<common::MouseEvent> mouseEvents;
+
   /// \brief Key event
   public: common::KeyEvent keyEvent;
 
+  /// \brief Max number of mouse events to process
+  public: const unsigned int kMaxMouseEventSize = 5u;
```
nkoenig, line 81: Curious about where this number came from

nkoenig (approved):
I had just one questions. Otherwise, works great.

src/plugins/minimal_scene/MinimalScene.cc
iche033: I picked a small number but it's still a little arbitrary. Main reason to keep a small queue size is that we don't want to keep too many mouse events in the queue otherwise we'll flood other gui plugins with mouse events that are potentially outdated. I expanded the doc in [f68805b](https://github.com/gazebosim/gz-gui/pull/502/commits/f68805bb7a207eb2b5b51806309db0c4eb43843a)

Issue #482, opened 2022-09-05
Can't open Nav Sat Map plugin (crashes)


ijnek:
## Environment
* OS Version: Ubuntu22.04
* Source or binary build? Binary (version: 6.6.1)

## Description
* Expected behavior: Nav Sat Map plugin opens correctly
* Actual behavior: Ignition Gazebo crashes with the following messages

```sh
[GUI] [Dbg] [Application.cc:426] Loading plugin [NavSatMap]
[GUI] [Wrn] [Application.cc:797] [QT] file::/NavSatMap/NavSatMap.qml:160:3: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... }
[GUI] [Wrn] [Application.cc:797] [QT] QObject::connect(QQuickWindow, QDeclarativeGeoMap): invalid nullptr parameter
[GUI] [Msg] Added plugin [Navigation satellite map] to main window
[GUI] [Msg] Loaded plugin [NavSatMap] from path [/usr/lib/x86_64-linux-gnu/ign-gui-6/plugins/libNavSatMap.so]
[GUI] [Wrn] [Application.cc:797] [QT] QGeoTileProviderOsm: Tileserver disabled at  QUrl("http://maps-redirect.qt.io/osm/5.8/satellite")
[GUI] [Wrn] [Application.cc:797] [QT] QGeoTileFetcherOsm: all providers resolved
```

## Steps to reproduce

1. ``ign gazebo -v 4``
2. Go to add a Nav Sat Map plugin from the right-top menu
3. Crashes, when Nav Sat Map plugin is clicked

## Output
![image](https://user-images.githubusercontent.com/18237662/188521559-3ad29718-1067-48bd-b0d4-1a999f76f891.png)

jennuine:
Hi @ijnek I have tried both fortress from source and binaries and am not able to reproduce the crash (I've tried launching `ign gazebo -v 4` then adding the plugin manually and also `ign gazebo -v 4 spherical_coordinates.sdf`). I suggest double checking your installation and make sure it's set up correctly.  An option is to uninstall then reinstall fortress: https://gazebosim.org/docs/fortress/install_ubuntu

jennuine:
I just realized you're on `jammy` not `focal`, re-opening issue since this wasn't tested appropriately

vincentrou:
[...]

PR #520, opened 2023-03-09
Switch to use mapboxgl map plugin in NavSat Map gui plugin

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-gui/issues/482

## Summary

Using the `osm` map plugin in the NavSat gui plugin seems to cause the window to freeze / crash. Switched to use `mapboxgl` instead.

Now it works for me:

navsat_mapboxgl

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/navsat_map/NavSatMap.qml
+++ b/src/plugins/navsat_map/NavSatMap.qml
@@ -91,7 +91,7 @@ Item {
 
   Plugin {
     id: mapPlugin
-    name: "osm"
+    name: "mapboxgl"
   }
 
   Map {


azeey:
From reading https://doc.qt.io/qt-5/location-plugin-mapboxgl.html, it says 

> Both Mapbox geo services plugins require an access token to access map styles and tiles hosted by Mapbox. 

and the documentation on `mabpoxgl.access_token` says

> Access token provided by Mapbox. The token can also be specified using the environment variable MAPBOX_ACCESS_TOKEN, but if also set using a plugin parameter, then this last one will have the precedence over the environment variable. When not set, a development token will be used by default. The development token is subject to the Mapbox Terms of Services and must not be used in production. This property has no effect on styles hosted outside the Mapbox servers.

So it's not clear to me if we're allowed to use it.

jennuine:
Also, when the box is moved using the transform tool, the GUI freezes up still :(

iche033:
hmm ok closing as it's not clear if we can use the plugin and it also does not completely fix the issue.

PR #429, opened 2022-06-28
Fix segfaulting tests

mjcarroll:
# 🦟 Bug fix

## Summary

This removes the deprecated scene3d tests for `garden`.  These have been a source of flakiness for quite some time in CI.

Additionally, this removes the `unload` calls for the rendering engines, which can cause issues.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #535, opened 2023-04-14
Fix tests by delaying get render engine calls 

iche033:
# 🦟 Bug fix

Targets #429

## Summary

Changes:
* moved `rendering::engine` calls to after the window had a chance to load the engine in tests. Otherwise the engine will be loaded in the main test thread and subsequent render calls will fail.
* reinstated calls to unload engine

## Checklist
- [ ] Added tests

### DIFF
--- a/test/integration/marker_manager.cc
+++ b/test/integration/marker_manager.cc
@@ -32,6 +32,7 @@
 #include <gz/utils/ExtraTestMacros.hh>
 
 #include "test_config.hh"  // NOLINT(build/include)
+#include "../helpers/TestHelper.hh"
 #include "gz/gui/Application.hh"
 #include "gz/gui/GuiEvents.hh"
 #include "gz/gui/MainWindow.hh"
@@ -142,18 +143,29 @@ TEST_F(MarkerManagerTestFixture,
   // Show, but don't exec, so we don't block
   window->QuickWindow()->show();
 
-  // Check scene
-  auto engine = rendering::engine("ogre2");
-  ASSERT_NE(nullptr, engine);
+  bool receivedRenderEvent{false};
+  auto testHelper = std::make_unique<TestHelper>();
+  testHelper->forwardEvent = [&](QEvent *_event)
+  {
+    if (_event->type() == events::Render::kType)
+    {
+      receivedRenderEvent = true;
+    }
+  };
 
   int sleep = 0;
   int maxSleep = 30;
-  while (0 == engine->SceneCount() && sleep < maxSleep)
+  while (!receivedRenderEvent && sleep < maxSleep)
   {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
     QCoreApplication::processEvents();
     sleep++;
   }
+  EXPECT_TRUE(receivedRenderEvent);
+
+  // Check scene
+  auto engine = rendering::engine("ogre2");
+  ASSERT_NE(nullptr, engine);
 
   EXPECT_EQ(1u, engine->SceneCount());
   scene = engine->SceneByName("scene");
--- a/test/integration/minimal_scene.cc
+++ b/test/integration/minimal_scene.cc
@@ -122,10 +122,6 @@ TEST(MinimalSceneTest, GZ_UTILS_TEST_ENABLED_ONLY_ON_LINUX(Config))
     }
   };
 
-  // Check scene
-  auto engine = rendering::engine("ogre2");
-  ASSERT_NE(nullptr, engine);
-
   int sleep = 0;
   int maxSleep = 30;
   while (!receivedRenderEvent && sleep < maxSleep)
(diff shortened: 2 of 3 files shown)


iche033:
hmm tests are fixed in focal but still failing on jammy

test/integration/marker_manager.cc
```diff
@@ -142,18 +143,29 @@ TEST_F(MarkerManagerTestFixture,
...
   {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
     QCoreApplication::processEvents();
     sleep++;
   }
+  EXPECT_TRUE(receivedRenderEvent);
+
+  // Check scene
+  auto engine = rendering::engine("ogre2");
+  ASSERT_NE(nullptr, engine);
```
azeey, line 168: Interesting. Since this pattern is useful in most tests here and the order of calls is important, could this be refactored into a `testing::getRenderEngine` function that blocks until the window is available and calles `rendering::engine` at the right time?

iche033: ok I added a helper function to do this. [acfcafc](https://github.com/gazebosim/gz-gui/pull/535/commits/acfcafc32af7024127e63add8c45eafe5e2e5731). One catch is that the function should only be called after the main window is visible. Added a doxygen comment on this.

azeey:
Looking at https://doc.qt.io/qt-5/qthread.html, it says

> It is important to remember that a QThread instance [lives in](https://doc.qt.io/qt-5/qobject.html#thread-affinity) the old thread that instantiated it, not in the new thread that calls [run](https://doc.qt.io/qt-5/qthread.html#run)(). This means that all of QThread's queued slots and [invoked methods](https://doc.qt.io/qt-5/qmetaobject.html#invokeMethod) will execute in the old thread. Thus, a developer who wishes to invoke slots in the new thread must use the worker-object approach; new slots should not be implemented directly into a subclassed QThread.

But our `RenderThread` class does implement a few slots even though it's a `QThread`. Should `RenderThread` be implemented using the worker-object approach?

iche033:
the implementation is based on an Qt example showing how to integrate opengl apps into QML. In that example, it also uses a QThread with slots. I can't seem to find the example any more so it could be outdated. There is this new example: https://doc.qt.io/qt-5/qtquick-scenegraph-openglunderqml-example.html. The approach is different. It maybe worth investigating the new approach later, e.g if we decide to upgrade to Qt6.

azeey:
Okay, I took a closer look and I see that our `RenderThread` class does not implement `run`, which means the default implementation will be used, which calls `QThread::exec` and starts the event loop for that object. I think it's essentially doing what the worker-object approach would be doing.

src/plugins/minimal_scene/MinimalScene.cc
```diff
@@ -687,6 +687,10 @@ void GzRenderer::Destroy()
 
     // TODO(anyone) If that was the last scene, terminate engine?
   }
+
+  // clean up in the rendering thread
```
azeey, line 691: I'm trying to understand why this fixes the issue. After `GzRenderer::Destroy` is called, the rendering thread exits, which means there should be no data races on `dataPtr->camera` and `dataPtr->rayQuery`, so it "should" be okay for them to be cleaned up in the destructor of `GzRenderer`. I feel like I'm missing something

iche033: In this PR, I think these 2 calls didn't make a difference but I have a habit of doing this now whenever I see it. 

I noticed that sometimes segfaults on jammy github actions come from the destructor of rendering objects as they try to invoke calls to destroy internal ogre/ogre2 resources. If the reset is not called here, the destructors are then called when the program exits (problem with shared ptr). By that time, the engine is already destroyed and the destructor would be trying to clean up internal resources that are no longer valid causing a segfault. In ogre2, we try to detect that as much as possible but we haven't caught everything yet since those segfaults still tend to happen on github actions (jammy only). 

Two other examples are https://github.com/gazebosim/gz-sensors/pull/324 
 and https://github.com/gazebosim/gz-sim/pull/1960 where I started reseting shared ptrs and that helped fixed a few failing tests on jammy.

The bigger problem here is the excessive use of shared pointer types in gz-rendering which will probably need to be addressed at some point.

azeey: hmm, so if destroying ogre2 resources after the engine is destroyed is problematic, is there a way to transfer ownership of those resources to the engine so it destroys them when it gets destroyed? I think Qt does that with its parent/child relationships. Maybe we can follow a similar approach. Anyways, thanks for the explanation.

azeey (approved):

src/plugins/minimal_scene/MinimalScene.cc
iche033: >  is there a way to transfer ownership of those resources to the engine so it destroys them when it gets destroyed?

yeah the engine actually does delete these ogre2 resources when it shuts down. In the destructor of rendering objects, we started adding checks to see if the engine is still loaded before attempting to request deletion again (by calling ogre2 destroy functions, which would cause a segfault if engine is already unloaded). I think we're just not very exhaustive in our checks. It difficult to pin point where the crash is coming from as well since it only happens on CI and not locally.

PR #558, opened 2023-07-28
Add message library generation to gui

mjcarroll:


CMakeLists.txt
```diff
@@ -68,6 +69,13 @@ set(GZ_TRANSPORT_VER ${gz-transport13_VERSION_MAJOR})
 gz_find_package(gz-rendering8 REQUIRED)
 set(GZ_RENDERING_VER ${gz-rendering8_VERSION_MAJOR})
 
+#--------------------------------------
+# Find gz-plugin
+gz_find_package(gz-sensors8-msgs REQUIRED)
+gz_find_package(gz-sensors8 REQUIRED)
```
azeey, line 75: Oof, do we need to depend on gz-sensors8 or is this just temporary?

proto/gz/msgs/air_pressure_sensor.proto
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+syntax = "proto3";
+package gz.msgs;
+option java_package = "com.gz.msgs";
+option java_outer_classname = "AirPressureSensorProtos";
```
azeey, line 21: hmm, I would have though this would go in `gz-sensors`.

mjcarroll: The `_sensor` messages are actually SDF-derived and not used by sensors, as far as I can tell.

PR #561, opened 2023-08-18
Fix rendering tests

iche033:
# 🦟 Bug fix

Targeted at https://github.com/gazebosim/gz-gui/pull/429

## Summary

The tests below are still failing on Jammy github actions in https://github.com/gazebosim/gz-gui/pull/429:
- INTEGRATION_transport_scene_manager
- INTEGRATION_minimal_scene

The failing tests were supposed to be fixed by https://github.com/gazebosim/gz-gui/pull/535 but for some reason they are now happening again.

The "fix" is more of a workaround. The main change is to not unload the render engine on exit which is what's causing the segfault in these 2 tests. It's safe to remove the `DestroyScene` call in the tests as that's taken care of by the minimal scene plugin. The tests do not fail on other CI builds and I'm also not able to reproduce them locally.

We could either get this in to make CI green or spend more time finding the root cause of the problem.

## Checklist
- [x] Added tests

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -691,15 +691,7 @@ void GzRenderer::Destroy()
 
     scene.reset();
     // TODO(anyone) If that was the last scene, terminate engine?
-    if (engine->SceneCount() == 0)
-    {
-      std::cerr << "terminating " << std::endl;
-      /// rendering::unloadEngine(engine->Name());
-      std::cerr << "terminated " << std::endl;
-    }
   }
-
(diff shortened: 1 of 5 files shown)


azeey (approved):
LGTM! Do you think it's worth figuring out why unloading the engine fails? For example, would it allow us to have multiple tests in the same file? Either way, I think we should get this in to fix the tests and create an issue for the root problem.

iche033:
> Do you think it's worth figuring out why unloading the engine fails?

I created https://github.com/gazebosim/gz-gui/issues/565 for this. I'll look into it as a background task.

PR #563, opened 2023-08-21
Make plugins include message types they use

mjcarroll:


src/plugins/point_cloud/PointCloud.cc
```diff
@@ -15,6 +15,7 @@
  *
  */
 
+#include "gz/msgs/marker.pb.h"
```
ahcorde, line 18: z::msgs::Float_V ?

src/plugins/camera_tracking/CameraTracking.cc
```diff
@@ -18,6 +18,7 @@
 #include <mutex>
 #include <string>
 
+#include <gz/msgs/gui_camera.pb.h>
```
ahcorde, line 21: msgs::Vector3d ?

ahcorde (changes requested):

ahcorde (approved):

PR #574, opened 2023-09-21
Fix crash when starting Topic Viewer

azeey:
# 🦟 Bug fix

## Summary

Starting the Topic Viewer in gz-sim causes a crash for me. The crash only seems to happen in one of my machines, so it might not be easy to reproduce. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/topic_viewer/TopicViewer.cc
+++ b/src/plugins/topic_viewer/TopicViewer.cc
@@ -202,8 +202,11 @@ void TopicViewerPrivate::CreateModel()
     std::vector<transport::MessagePublisher> publishers;
     std::vector<transport::MessagePublisher> subscribers;
     this->node.TopicInfo(topics[i], publishers, subscribers);
-    std::string msgType = publishers[0].MsgTypeName();
-    this->AddTopic(topics[i], msgType);
+    if (!publishers.empty())
+    {
+      std::string msgType = publishers[0].MsgTypeName();
+      this->AddTopic(topics[i], msgType);
+    }
   }
 }
 


src/plugins/topic_viewer/TopicViewer.cc
```diff
@@ -202,8 +202,11 @@ void TopicViewerPrivate::CreateModel()
     std::vector<transport::MessagePublisher> publishers;
     std::vector<transport::MessagePublisher> subscribers;
     this->node.TopicInfo(topics[i], publishers, subscribers);
-    std::string msgType = publishers[0].MsgTypeName();
-    this->AddTopic(topics[i], msgType);
+    if (!publishers.empty())
```
azeey, line 205: @caguero I'm not sure how this condition (empty `publishers`) could occur if the number of topics returned by `TopicList` is non-zero, but it actually does. I also see a similar logic used in `UpdateModel`, so this doesn't seem to be a new issue.

mjcarroll: Is it now returning subscription-only topics?

caguero: Exactly, `TopicList()` now returns publishers and subscribers. So if you have a single subscriber for a given topic (and not a publisher), you shouldn't do `publishers[0]` because it'll crash.

azeey: Okay, we need to update the documentation then. It still says it returns advertised topics. We should also mention this in `Migration.md` since it's somewhat of a breaking change.

iche033 (approved):

PR #582, opened 2023-10-05
Tidy namespaces

mjcarroll:
Housekeeping to clean up namespaces.

Remove `using namespace gz` and use nested namespace declarations where possible.

### DIFF
--- a/include/gz/gui/Application.hh
+++ b/include/gz/gui/Application.hh
@@ -37,191 +37,188 @@ namespace tinyxml2
   class XMLElement;
 }
 
-namespace gz
+namespace gz::gui
 {
-  namespace gui
-  {
-    class ApplicationPrivate;
-    class Dialog;
-    class MainWindow;
-    class Plugin;
+class ApplicationPrivate;
+class Dialog;
+class MainWindow;
+class Plugin;
 
-    /// \brief Type of window which the application will display
-    enum class WindowType : int
-    {
-      /// \brief A main window, which may contain top-level menus and multiple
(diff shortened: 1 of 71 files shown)


mjcarroll:

include/gz/gui/Application.hh
```diff
@@ -37,191 +37,188 @@ namespace tinyxml2
   class XMLElement;
 }
 
-namespace gz
+namespace gz::gui
```
jennuine, line 40: Do we plan to update all the libs within the gazebo ecosystem? If not, I'm slightly against making these changes so that we can keep all the libs in similar style

mjcarroll: Aspirationally, I would like to.  Realistically, it takes time.  I think that the more condensed looks and feels better, so moving in that direction over time seems appealing.

jennuine: Is there a reason we aren't starting at gz-gui7 or earlier? Will this make forward ports more painful?

mjcarroll: I think we need at least C++17, so targeting gui7 would be reasonable.

jennuine: let's do that. Also, we should create an issue that tracks the changes for all the other libs with a 'help wanted' label. It could be easy tasks for outside contributors

mjcarroll: Just opening another PR seems easier: https://github.com/gazebosim/gz-gui/pull/590

PR #584, opened 2023-10-06
First pass at level1 clazy checks for qt

mjcarroll:
Cleaning up `clazy` warnings to make sure we are using Qt correctly.

### DIFF
--- a/src/MainWindow_TEST.cc
+++ b/src/MainWindow_TEST.cc
@@ -342,7 +342,7 @@ TEST(MainWindowTest,
 
   // Access window after it's open
   bool closed{false};
-  QTimer::singleShot(300, [&closed]
+  QTimer::singleShot(300, App(), [&closed]
   {
     auto win = App()->findChild<MainWindow *>();
     ASSERT_NE(nullptr, win);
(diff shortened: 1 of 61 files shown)


src/MainWindow_TEST.cc
```diff
@@ -342,7 +342,7 @@ TEST(MainWindowTest,
 
   // Access window after it's open
   bool closed{false};
-  QTimer::singleShot(300, [&closed]
+  QTimer::singleShot(300, App(), [&closed]
   {
     auto win = App()->findChild<MainWindow *>();
```
jennuine, line 347: should this be:
```suggestion
  auto *app = App();
  QTimer::singleShot(300, app, [&closed]
  {
    auto *win = app->findChild<MainWindow *>();
```
?

src/MainWindow_TEST.cc
```diff
@@ -519,7 +519,10 @@ void FindExitDialogButtons(
   ASSERT_NE(nullptr, dialog);
 
   QObject *buttonBox{nullptr};
-  for (const auto& c : dialog->findChildren<QObject *>())
+  // Qt considers range-based for loops over temporary objects
+  // potentially dangerous, so explicitly create a container.
+  const auto children = dialog->findChildren<QObject *>();
```
jennuine, line 524:
```suggestion
  const auto *children = dialog->findChildren<QObject *>();
```

jennuine (approved):

src/MainWindow_TEST.cc
mjcarroll: Yes, I went ahead an addressed the rest of the places where we do this.

src/MainWindow_TEST.cc
mjcarroll: [273fac0](https://github.com/gazebosim/gz-gui/pull/584/commits/273fac06c0698033a9ee5db13eb296e791139532)

PR #590, opened 2023-10-30
Tidy namespaces

mjcarroll:
Switch to nested namespaces, as they are supported now.

Replaces https://github.com/gazebosim/gz-gui/pull/582

Split into three commits for easy review:

* Main code: https://github.com/gazebosim/gz-gui/pull/590/commits/6b2a501598f8684eb328db6b3c3ed0b83769f954
* Plugins: https://github.com/gazebosim/gz-gui/pull/590/commits/ccdef7452953f229ea75fe9dc545e26a38c164cf
* Whitespace https://github.com/gazebosim/gz-gui/pull/590/commits/b939f2eb1ca3996ed529b139018a1ebc4580af5e


include/gz/gui/Application.hh
```diff
@@ -37,188 +37,195 @@ namespace tinyxml2
...
-      private: std::unique_ptr<ApplicationPrivate> dataPtr;
-    };
-
-    /// \brief Get current running application, this is a cast of qGuiApp.
-    /// \return Pointer to running application, or nullptr if none is running.
-    GZ_GUI_VISIBLE
-    Application *App();
-  }
-}
+class ApplicationPrivate;
```
ahcorde, line 42: can we include these classes instead of this forward declaration ?

ahcorde (changes requested):

include/gz/gui/Application.hh
mjcarroll: I believe that at least `ApplicationPrivate` doesn't have a declaration in the public API, so we wouldn't be able to.

azeey (approved):
One comment which should hopefully fix CI.

examples/standalone/custom_drawer/custom_drawer.cc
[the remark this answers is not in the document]
mjcarroll: Yep, I must have missed that one.

ahcorde (approved):

PR #594, opened 2023-11-09
Added motion duration to the 'move to pose' service of the camera tracking plugin.

jrutgeer:
# 🎉 New feature

This PR adds the possibility to provide the duration time to the 'move to pose' service of the camera tracking plugin.

If no or negative duration is specified, the duration defaults to the previous value of 0.5 seconds.


azeey:
This needs https://github.com/gazebosim/gz-msgs/pull/408 and since that needs to target `main`, this should also target `main`.

jrutgeer:
@azeey I retargetted to main. It seems no further steps (similar to [these](https://github.com/gazebosim/gz-msgs/pull/408#issuecomment-1812058816)) are necessary as the 'Files changed' show only changes related to this PR?

ahcorde:
This one looks fine

ahcorde (changes requested):
can you fix the conflicts ?

jrutgeer:
There seems to be only one issue, which is due to gazebosim/gz-msgs#408 being required for this to compile?

```
/github/workspace/src/plugins/camera_tracking/CameraTracking.cc:296:12: error: 'const class gz::msgs::GUICamera' has no member named 'duration'
    296 |   if (_msg.duration() > 0)
        |            ^~~~~~~~
```

ahcorde:
I was refering to this conflicts 

aconflicts

jrutgeer:
No, I hadn't seen that.

To my defence:
For a regular contributor like me, for whom this isn't part of the day to day job, it is **very** hard to see the forest for the trees in this whole process:
- Not clear which branch one should target,
- No feedback upon change of base branch from `gz-gui8` to `main` that `main` was changed in the mean time wrt the `main` in my fork,
- Resolve requests in the web gui:
  - It wasn't clear to me where that came from (as locally the change compiled correctly),
  - Resolve through the web gui is obviously prone to errors, as you can't compile and check,

So for this last commit (hopefully!), I set up a local source tree targetting main:
- There are no instructions for this (that I am aware of),
- So I had taken `collection-harmonic.yaml` file and changed each version to those mentioned in the `CMakeLists.txt` files (e.g. `gz-common5` to `gz-common6`),
- Only to find out `gz-common6` branch does not exits, it should just be `main`,

Obviously in hindsight this all makes sense, but... what a learning curve in order to supply such a simple patch.


Imo. the ROS procedure, where you just work with a local `rolling` install and provide PR's wrt that, and `rolling` is always up to date with all latest patches, is **much** more straight-forward.

I asked about the Gazebo Sim way of working [here](https://robotics.stackexchange.com/questions/24863/what-is-the-equivalent-to-rolling-for-gazebo-sim) but I still have no true understanding the best-practice procedure to submit features and bug fixes for Gazebo Sim.

jennuine:
@jrutgeer we appreciate your feedback and your contribution!

Here has been my workflow in case it helps, I have a workspace for the distribution I'm working on. If there is a breaking ABI change in any of the libraries, then the feature needs to land in `main`. The next major release will be Ionic and its collection of libraries can be found here: https://github.com/gazebo-tooling/gazebodistro/blob/master/collection-ionic.yaml

When the next version is released, we also announce the name of the version. For example, when we released Harmonic we announced the release after that will be Ionic. The yaml collection is subject to change when we decide to bump a library version so it's good to be aware of changes and update the new version workspace when possible. In Ionic's case we are bumping most of the libraries (except for `gz-tools`), which is why they are almost all targeting `main`. In the future if the gazebo team sees there there needs to be breaking changes in `gz-tools` then it will be bumped to use `main` as well.

I hope this can provide a little more clarity. We apologize for the confusion but are doing our best with the time and resources we have allocated. Any contributions are appreciated and welcome 😄

PR #629, opened 2024-07-24
Fix banding and discoloration issue in 3D scene

iche033:
# 🦟 Bug fix

## Summary

The MinimalScene uses the `GammaAdjust` effect for gamma correction otherwise the scene will appear darker than it should when using ogre2 + OpenGL. However, it was found that caused caused banding and discoloration issues in the 3D scene, see screenshots [here](https://github.com/azeey/gz-rendering-benchmark/tree/0c670ced35a664558fad760ca428a19dd94d68a4/gamma_correction). This PR removes the `GammaAdjust` QML effect and instead applies a workaround in the OpenGL RHI to convert the input ogre2 render texture from the sRGB format to linear RGB format before passing it to Qt. The scene now display the correct colors but without banding issues.

Note this may not be necessary in Qt 6 as I see that there is a new class and API to [set the flags](https://doc.qt.io/qt-6/qrhitexture.html#Flag-enum) of its RHI texture to sRGB color space

## Checklist
- [ ] Added tests


jennuine (approved):
Looks good, whenever we migrate to Qt6 we'll need to remember to try w/o the workaround before we forward port the code. Thanks Ian.

iche033:
replaced by https://github.com/gazebosim/gz-gui/pull/630

Issue #586, opened 2023-10-10
Migrate Qt5 to Qt6


mjcarroll:
Tracking issue for tasks required to migrate from Qt5 to Qt6.

- [ ] Update MinimalScene for new APIs getting native access to VkImage and Metal equivalent.
- [ ] QtGraphicalEffects is deprecated and has some incompatibilities: https://doc.qt.io/qt-6/qtgraphicaleffects5-index.html

mjcarroll:
Clazy checks: https://github.com/gazebosim/gz-gui/pull/584

mjcarroll:
Bulk of the migrations for gui: https://github.com/gazebosim/gz-gui/pull/596

azeey:
Here's the list of qt6 packages I needed to install:

- qml6-module-qt-labs-folderlistmodel
- qml6-module-qt-labs-platform
- qml6-module-qt-labs-settings
- qml6-module-qtcharts
- qml6-module-qtpositioning
- qml6-module-qtqml-models
- qml6-module-qtquick-controls
- qml6-module-qtquick-dialogs
- qml6-module-qtquick-layouts
- qml6-module-qtquick-templates
- qml6-module-qtquick-window
- qml6-module-qtquick
- qt6-base-dev
- qt6-declarative-dev
- libqt6quickcontrols2-6
- qml6-module-qtcore
- qml6-module-qtqml-workerscript
- qml6-module-qt5compat-graphicaleffects
- libqt6svg6

azeey:
I've added a fix in the gz-sim branch https://github.com/gazebosim/gz-sim/pull/2305/commits/a2652ac19b906587f441c40f29c886ceda7dd0d3 which should help with the communication between the server and GUI ECMs. With that, I see glimmers of hope :)
[Screen recording 2024-07-16 14.18.43.webm](https://github.com/user-attachments/assets/8281d76d-4bce-46b8-8ffd-445ec0aebaaa)

caguero:
I was able to visualize the 3D screen even without your changes. It was hidden at the very left covered by other collapsed window plugins.

![Screenshot from 2024-07-17 16-51-50](https://github.com/user-attachments/assets/da205744-47a1-4a2c-99bb-34391434bea8)

azeey:
Implemented by https://github.com/gazebosim/gz-sim/pull/2832, https://github.com/gazebosim/gz-gui/pull/666 and a few other auxiliary PRs.

PR #651, opened 2025-01-29
Use SplitView from QtQuick.Controls 2 for GzSplit

azeey:
# 🎉 New feature

Toward #586 

## Summary
The SplitView from QtQuick.Controls 1 has been deprecated for a while and has been removed from Qt6. This change is a stepping stone for migrating to Qt6.

The new SplitView is significantly different from the previous one and so major changes had to be made. The original GzSplit component was also designed with more capabilities in mind for the future, which complicated the code. This has been simplified:

* Instead of determining whether to add a new split for each card added, two splits are hard coded: (1) the main view port (2) the side panel. For every new card added, the code simply checks adds the card to the main view port if its empty. Otherwise, it puts it in the side panel.

* Animations have been removed. It was difficult to get consistent behavior with different split sizes.

* Card items are returned instead of their names when calling `addCard`. This simplifies the code that searches for items by name

* Logic to detect if cards have been collapsed manually by resizing the card has been removed. It's not clear if it is even possible to do this with any of the cards/plugins in Gazebo since the `minimumHeight` constraint would prevent resizing the cards that small.

**New SplitView**
image

**Old SplitView**
image

## Test it

- `gz gui -c examples/config/layout.config`
- `gz sim camera_sensor.sdf`

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/gui/qml/GzCard.qml
+++ b/include/gz/gui/qml/GzCard.qml
@@ -15,8 +15,7 @@
  *
 */
 import QtQuick 2.9
-import QtQuick.Controls 1.4 as QQC1
-import QtQuick.Controls 2.2
+import QtQuick.Controls 2.15
 import QtQuick.Controls.Material 2.1
 import QtQuick.Layouts 1.3
 import QtQuick.Window 2.2
@@ -24,6 +23,7 @@ import "qrc:/qml"
 
 // TODO: don't use "parent"
 Pane {
+  clip: true
   /**
    * Minimum length of each dimension
    */
@@ -95,16 +95,6 @@ Pane {
   property string cardBackground: "#00000000"
 
   /**
-   *
-   */
-  property var backgroundItem: null
-
-  /**
-   * Stores last height of plugin to expand to.
-   */
-  property int lastHeight: 50
-
-  /**
    * True if there's at least one anchor set for the card.
    * There's no way to check the anchors themselves, so we need
    * to keep track of this ourselves.
@@ -112,14 +102,23 @@ Pane {
   property bool anchored: false
 
   /**
+   * Keep track of whether the card is floating
+   */
+  property bool floating: false
+
+  /**
    * Minimum width of the card pane
    */
-  property int cardMinimumWidth: 250;
+  property alias cardMinimumWidth: content.minimumWidth;
 
   /**
    * Minimum height of the card pane
    */
-  property int cardMinimumHeight: 250;
+  property double cardMinimumHeight: content.minimumHeight;
+
+
+  property double cardMaximumHeight: Infinity
+
 
   /**
    * Tool bar background color
@@ -158,6 +157,10 @@ Pane {
    */
   objectName: "plugin" + Math.floor(Math.random() * 100000);
 
+
+  contentWidth: content.implicitWidth
+  contentHeight: content.implicitHeight
+
   // Stop scroll propagation to widgets below
   MouseArea {
     anchors.fill: parent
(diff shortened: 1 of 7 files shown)


include/gz/gui/qml/GzCard.qml
```diff
@@ -94,32 +94,31 @@ Pane {
...
+  /**
+   * Keep track of whether the card is floating
+   */
+  property bool floating: false
+
   /**
    * Minimum width of the card pane
    */
-  property int cardMinimumWidth: 250;
+  property alias cardMinimumWidth: content.minimumWidth;
```
iche033, line 112: I think this we need to update the test for this change:

```
  /github/workspace/src/Plugin_TEST.cc:304: Failure
  Expected equality of these values:
    propElem->Attribute("type")
      Which is: "double"
    pluginTypes[propElem->Attribute("key")]
      Which is: "int"
```

iche033 (commented):
Nice! works great for me. Not sure if it's possible to make the split view handle thinner or change its color to gray but those are minor details that can come later.

include/gz/gui/qml/GzCard.qml
azeey: Fixed in https://github.com/gazebosim/gz-gui/pull/651/commits/2570a75ba26a51926f7c07cd0adf6804ee2b899e

I opted to ignore `card*` properties from being parsed from XML since that wasn't working anyway. Looking at the implementation, I don't think they were intended to be exposed since they are updated internally to collapse/expand the card.

src/Application.cc
```diff
@@ -737,22 +730,12 @@ bool Application::AddPluginsToWindow()
...
-             << splitName.toString().toStdString() << "]" << std::endl;
-      return false;
-    }
+    // Add card to GzSplit
+    QMetaObject::invokeMethod(bgItem, "addCard",
+        Q_ARG(QVariant, QVariant::fromValue(cardItem)));
 
     // Add card to main window
-    cardItem->setParentItem(splitItem);
+    // gzdbg << "Adding to:" << splitItem.toString().toStdString() << std::endl;
```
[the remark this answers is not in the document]
azeey: Removed.

azeey:
> Nice! works great for me. Not sure if it's possible to make the split view handle thinner or change its color to gray but those are minor details that can come later.

It's possible to change the color and make it thinner, but it causes the mouse area to become thinner as well, which makes it really hard to use. The previous version had a wider mouse area with a thin (1px wide) rectangle display. With the SplitView in QtQuick.Control2, it's not possible to control the mouse area and the displayed rectangle independently.

https://github.com/gazebosim/gz-gui/pull/651/commits/2570a75ba26a51926f7c07cd0adf6804ee2b899e also fixes a failing test in `UNIT_MainWindow_TEST`. The test was requesting to resize the window to a size below the minimum size set in https://github.com/gazebosim/gz-gui/blob/2570a75ba26a51926f7c07cd0adf6804ee2b899e/include/gz/gui/qml/Main.qml#L30-L31, so I'm not sure how it was working in the first place.

iche033:
the noble build picked up some weird warnings:

```
[GNU C Compiler (gcc)] [-ERROR-] - '/home/jenkins/workspace/gz_gui-ci-pr_any-noble-amd64/build/file:/usr/lib/x86_64-linux-gnu/qt5/qml/QtQuick/Controls.2/Material/SplitView.qml' file not found
```

azeey:
> the noble build picked up some weird warnings:

Not sure wht's going on. I can't reproduce this locally. I'm going to try to retrigger the job to see if it happens consistently.

azeey:
> It's possible to change the color and make it thinner, but it causes the mouse area to become thinner as well, which makes it really hard to use. 

There's a bug reported on this (https://bugreports.qt.io/browse/QTBUG-122929), and there's a workaround in qt6, so will take a look again after we migrate.

azeey:
> the noble build picked up some weird warnings:

Fixed in https://github.com/gazebosim/gz-gui/pull/651/commits/c5ea53f2b3e92122d9bf4f23e503c4d03408886d

iche033 (approved):

PR #656, opened 2025-02-26
Fix GzSpinBox in cases where the value property has a binding

azeey:
# 🦟 Bug fix

## Summary
Before this fix, if the value property has a binding instead of a value,
any modification from the GUI (e.g. hitting the up arrow on the spin
box) would not actually change the displayed value because it is
immediately reset back to the value set by the binding. This is the case
in the `GzPose` module and the effect can be easily be seen in the
`GridConfig` plugin. To test, run:

`gz gui -c examples/config/grid_config.config`

You will see that, before this change, modifying the x,y,z,roll,pitch,
yaw values actually changes the grid configuration, but the updated values are
not displayed in the spinbox.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/gui/qml/GzSpinBox.qml
+++ b/include/gz/gui/qml/GzSpinBox.qml
@@ -29,6 +29,7 @@ Item {
   property int decimals: 0
 
   onValueChanged: {
+    console.log("root val", value)
     if (!spinBox.__modifying)
     {
       spinBox.value = decimalToInt(root.value)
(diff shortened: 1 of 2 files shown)


iche033 (approved):
This works for me. While testing, I ran into another issue with setting grid pose in GridConfig plugin. I'll create a separate PR for this.

iche033:
> This works for me. While testing, I ran into another issue with setting grid pose in GridConfig plugin. I'll create a separate PR for this.

https://github.com/gazebosim/gz-gui/pull/658

PR #596, opened 2023-11-10
Qt6 WIP

mjcarroll:

PR #657, opened 2025-02-26
Migrate to Qt6

azeey:
# 🎉 New feature

Closes https://github.com/gazebosim/gz-gui/issues/586

Requires #656 

## Summary

Based on https://github.com/gazebosim/gz-gui/pull/596, but offers a simpler migration. Mainly, we keep using the plugin approach we were using before and set context properties to make the plugin's functionality available in QML modules. The plugin's name is prepended with an underscore (_) since properties that start with an uppercase letter are not allowed in Qt6. See https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html#defining-property-attributes

> Property names must begin with a lower case letter and can only contain letters, numbers and underscores.

This PR also does not try to support both Qt5 and Qt6 as it's not clear right now if that will actually be possible. We can revisit this later if we need to.

### Notes
- Not all imports need to have their versions removed, but the following do:
  - `QtQuick.Dialogs`
  - `Qt5Compat.GraphicalEffects`
  - `QtCore`

## TODO:
- [ ] Fix commented out code
- [ ] Fix TreeView implementation
- [ ] Check that the changes to `FileDialog` loses functionality
- [ ] NavSatMap:  QtLocation is not available in Qt6 6.4 which is what's shipped in Ubuntu Noble. It is available in newer versions of Ubuntu https://packages.ubuntu.com/oracular/qml6-module-qtlocation.

## Checklist
- [ ] Added tests


iche033:
replaced by #666

PR #669, opened 2025-03-30
Image Display Depth Visual Flip

exMachina316:
# 🎉 New feature

Closes https://github.com/gazebosim/gz-sensors/issues/473

### Issue brief

**Expected behavior:**
The rviz2 visualizer for depth image should match the visualization in the Image Display Viewer in Gazebo

**Actual behavior:**
The far point of a depth image in rviz2 is set as white but in gazebo it is black. These are default behaviors for these systems for a long time. 

| Rviz2    | Gazebo Image Display |
| ---------- | ---------------------------------|
| image | image |

### Solution
To quote @azeey 
> There doesn't seem to be a standard on how to display depth images. A quick look on the web shows both approaches being used in different places. Since the behavior in Gazebo has been around for a long time, I don't think we should change it. Perhaps a flag in the "Image Display" plugin to revert the colors could be an option. https://github.com/gazebosim/gz-sensors/issues/473#issuecomment-2457773426

From this, I simply added a checkbox in the image display plugin which switches a 'flip/no flip' flag in qml and a callback function to handle the toggling. This flag is passed directly into `common::Image::ConvertToRGBImage` as the `_flip` argument for `float32` type image, where the actual pixel value flipping is handled. This argument was previously set to `true` always, causing the mismatch.
**Note:** A new mutex was also introduced named `service_lock` to safely update the flag.

### Unit Tests
Implemented a couple of tests to ensure (in order):
1. The object can be found by its name.
2. The checkbox property is valid, i.e. is loaded properly.
3. The checkbox property matches the default value set in the code on startup.

## Manually test it
1. Launch a sdf with a depth camera using ros2 and bridge.
4. Add the Image Display plugiin gazebo.
5. View the depth image in rviz2.
6. Notice that far point in gazebo is black but in rviz it is white. This is the mismatch issue.

| Rviz2    | Gazebo Image Display |
| ---------- | ---------------------------------|
|Screenshot from 2025-03-30 13-41-06 | Screenshot from 2025-03-30 13-41-15 |  

7. Click the Flip Depth Visual checkbox in Image Display window, and watch the rviz2 image and gazebo image match.

| Rviz2    | Flipped Gazebo Image Display |
| ---------- | ---------------------------------|
|Screenshot from 2025-03-30 13-41-06 | Screenshot from 2025-03-30 13-41-19 |

## Checklist
- [x] Added tests

### DIFF
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -23,6 +23,7 @@
 #include <limits>
 #include <string>
 #include <vector>
+#include <mutex>
 
 #include <gz/common/Console.hh>
 #include <gz/common/Image.hh>
@@ -49,8 +50,15 @@ class ImageDisplay::Implementation
   /// \brief Mutex for accessing image data
   public: std::recursive_mutex imageMutex;
 
+  /// \brief Mutex for variable mutated by the checkbox.
+  /// The variables are: flipDepthVisual
+  public: std::mutex serviceMutex;
+
   /// \brief To provide images for QML.
   public: ImageProvider *provider{nullptr};
+
+  /// \brief Flip visual flag
+  public: bool flipDepthVisual{true};
 
   /// \brief Holds the provider name unique to this plugin instance
   public: QString providerName;
@@ -108,6 +116,8 @@ void ImageDisplay::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
 
   this->PluginItem()->setProperty("showPicker", topicPicker);
 
+  this->dataPtr->flipDepthVisual = true;
+
   if (!topic.empty())
   {
     this->SetTopicList({QString::fromStdString(topic)});
@@ -120,6 +130,7 @@ void ImageDisplay::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
 void ImageDisplay::ProcessImage()
 {
   std::lock_guard<std::recursive_mutex> lock(this->dataPtr->imageMutex);
+  std::lock_guard<std::mutex> service_lock(this->dataPtr->serviceMutex);
 
   unsigned int height = this->dataPtr->imageMsg.height();
   unsigned int width = this->dataPtr->imageMsg.width();
@@ -142,7 +153,8 @@ void ImageDisplay::ProcessImage()
       // i.e. darker pixels = higher values and brighter pixels = lower values
       common::Image::ConvertToRGBImage<float>(
           this->dataPtr->imageMsg.data().c_str(), width, height, output,
-          0.0f, std::numeric_limits<float>::lowest(), true);
+          0.0f, std::numeric_limits<float>::lowest(),
+          this->dataPtr->flipDepthVisual);
       break;
     case msgs::PixelFormatType::L_INT16:
       common::Image::ConvertToRGBImage<uint16_t>(
@@ -193,6 +205,15 @@ void ImageDisplay::ProcessImage()
 
   this->dataPtr->provider->SetImage(image);
   emit this->newImage();
+}
+
+//////////////////////////////////////////////////
+void ImageDisplay::FlipDepthVisual(bool _value)
+{
+  std::lock_guard<std::mutex> service_lock(this->dataPtr->serviceMutex);
+  this->dataPtr->flipDepthVisual = !_value;
+  gzmsg << "Depth Visualization " << ((_value) ? "Flipped." : "Standard.")
+        << std::endl;
 }
 
 /////////////////////////////////////////////////
--- a/src/plugins/image_display/ImageDisplay.cc
+++ b/src/plugins/image_display/ImageDisplay.cc
@@ -51,14 +51,14 @@ class ImageDisplay::Implementation
   public: std::recursive_mutex imageMutex;
 
   /// \brief Mutex for variable mutated by the checkbox.
-  /// The variables are: flipDepthVisual
+  /// The variables are: flipDepthVisualization
   public: std::mutex serviceMutex;
 
   /// \brief To provide images for QML.
   public: ImageProvider *provider{nullptr};
 
-  /// \brief Flip visual flag
-  public: bool flipDepthVisual{true};
+  /// \brief Flip visualization flag
+  public: bool flipDepthVisualization{true};
 
   /// \brief Holds the provider name unique to this plugin instance
   public: QString providerName;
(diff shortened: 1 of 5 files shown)


src/plugins/image_display/ImageDisplay.cc
```diff
@@ -49,9 +49,16 @@ class ImageDisplay::Implementation
   /// \brief Mutex for accessing image data
   public: std::recursive_mutex imageMutex;
 
+  /// \brief Mutex for variable mutated by the checkbox.
+  /// The variables are: flipDepthVisual
+  public: std::mutex serviceMutex;
```
ahcorde, line 54: include `<mutex>`

ahcorde (changes requested):

ahcorde (changes requested):
[Linters are failing](https://github.com/gazebosim/gz-gui/actions/runs/14167760268/job/39699735552?pr=669)

```bash
  /github/workspace/src/plugins/image_display/ImageDisplay_TEST.cc:99:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
  /github/workspace/src/plugins/image_display/ImageDisplay_TEST.cc:102:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
  /github/workspace/src/plugins/image_display/ImageDisplay_TEST.cc:103:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```

src/plugins/image_display/ImageDisplay.qml
```diff
@@ -58,6 +58,19 @@ Rectangle {
...
+      Layout.alignment: Qt.AlignHCenter
+      id: displayVisual
+      Layout.columnSpan: 6
+      Layout.fillWidth: true
+      text: qsTr("Flip Depth Visual")
+      checked: false
+      onClicked: {
+        ImageDisplay.FlipDepthVisual(checked);
+      }
+    }
```
iche033, line 72: consider adding a property that lets user configure whether or not to display this `CheckBox`? Same idea as the how `showPicker` property is used to determine the visibility of the topic picker below. 

Mainly because the `sensors_demo.sdf` demo world now has this checkbox shown for all non-depth image displays as well. It would be nice to hide them

src/plugins/image_display/ImageDisplay.qml
```diff
@@ -58,6 +58,19 @@ Rectangle {
     anchors.fill: parent
     anchors.margins: 10
 
+    CheckBox {
+      objectName: "flipDepthImageColorCheckBox"
+      Layout.alignment: Qt.AlignHCenter
+      id: displayVisual
+      Layout.columnSpan: 6
+      Layout.fillWidth: true
+      text: qsTr("Flip Depth Visual")
```
iche033, line 67: nit: Visual in Gazebo means a slight different meaning than visualization. 

```suggestion
      text: qsTr("Flip Depth Visualization")
```

src/plugins/image_display/ImageDisplay.qml
```diff
@@ -58,6 +58,19 @@ Rectangle {
     anchors.fill: parent
     anchors.margins: 10
 
+    CheckBox {
```
iche033, line 61: add tooltip?

iche033 (commented):
thanks for the contribution! I added some minor comments but this looks great.

src/plugins/image_display/ImageDisplay.qml
exMachina316: Yes I understand, it sounds good. Additionally we could auto hide it too for non-depth image topics, although the sudden changes in UI when the image changes might look clunky and also there is the handling of this property from the config which has to have more precedence over the topic type change trigger.

ahcorde (approved):

src/plugins/image_display/ImageDisplay.cc
```diff
@@ -142,7 +162,8 @@ void ImageDisplay::ProcessImage()
       // i.e. darker pixels = higher values and brighter pixels = lower values
       common::Image::ConvertToRGBImage<float>(
           this->dataPtr->imageMsg.data().c_str(), width, height, output,
-          0.0f, std::numeric_limits<float>::lowest(), true);
+          0.0f, std::numeric_limits<float>::lowest(),
+          this->dataPtr->flipDepthVisualization);
```
iche033, line 166: by default `this->dataPtr->flipDepthVisualization` is initialized to `true`, meaning we show depth image as flipped right? However, the checkbox is unchecked when the image display plugin loaded. I think we should bind the checkbox value to the `this->dataPtr->flipDepthVisualization` variable so that it's checked on startup?

src/plugins/image_display/ImageDisplay.qml
```diff
@@ -58,6 +63,24 @@ Rectangle {
...
+      Layout.fillWidth: true
+      text: qsTr("Flip Depth Visualization")
+      checked: false
+      onClicked: {
+        ImageDisplay.SetFlipDepthVisualization(checked);
+      }
+      ToolTip.visible: hovered
+      ToolTip.delay: tooltipDelay
+      ToolTip.timeout: tooltipTimeout
+      ToolTip.text: qsTr("Flip the depth image color to match rviz")
```
iche033, line 81: nit: I think we can omit mentioning rviz here

```suggestion
      ToolTip.text: qsTr("Flip the depth image color")
```

src/plugins/image_display/ImageDisplay.cc
exMachina316: I kept the checkbox behavior and the internal flip flag logically inverse of each other i.e.

    this->dataPtr->flipDepthVisualization = !_value;    // _value holds the checked status of the UI element

 This give the illusion of the default behavior being the same without any new auto selected settings popping up. Should I change it so that it is auto-selected and the relationship between them be logically equal?

src/plugins/image_display/ImageDisplay.qml
exMachina316: @iche033 should I add the auto hide feature or keep the current 'config' behavior added in the latest commit?

src/plugins/image_display/ImageDisplay.cc
iche033: oh ok I see. Yeah I think it would be good to keep the relationship logically equal

iche033 (approved):

iche033:
latest changes look good to me, merging!

PR #676, opened 2025-04-18
Remove build RUNPATH from installed libGrid3D.so

### LINKED ISSUE #627: libGrid3D.so is installed with RUNPATH set to the build location
## Environment
* OS Version: NixOS (not relevant)
* Source or binary build? Both (branch `gz=gui8`)
  
  

## Description
* Expected behavior: `libGrid3D.so` is not installed with `RUNPATH` set.
* Actual behavior: `libGrid3D.so` is has `RUNPATH` set.

## Steps to reproduce

Run the following and observe the output:

```
$ objdump -x /usr/lib/x86_64-linux-gnu/gz-gui-8/plugins/libGrid3D.so|grep PATH
RUNPATH              /home/jenkins/workspace/gz-gui8-debbuilder/build/gz-gui-8.3.0/obj-x86_64-linux-gnu/lib:
$ objdump -x /usr/lib/x86_64-linux-gnu/gz-gui-8/plugins/libGridConfig.so|grep PATH
# no output
```
`libGrid3D.so` has RUNPATH set, while other plugins don't.

## Discussion

I came across this problem when building Gazebo with the Nix package manager in [this project](https://github.com/lopsided98/nix-ros-overlay). Nix builds are automatically checked for inappropriate RUNPATH and fail if one is detected.

The reason why `RUNPATH` is set for `libGrid3D.so` is that it just a copy of `libGridConfig.so`, but installed differently ([see here](https://github.com/gazebosim/gz-gui/blob/036df090d5fb9323617ad186156e295a85e38421/src/plugins/grid_config/CMakeLists.txt#L11)). Due to the fact that it is installed as a file rather than a library, CMake doesn't remove `RUNPATH` setting during installation.

I tried to figure out, how to convince CMake to install a library under a different name and remove `RUNPATH` settings, however I didn't find anything straightforward (except building the library twice). `RENAME` keyword is ignored when installing library targets and I don't see a way to do post-installation changes.

We can work around this in the downstream project, but fixing it here might benefit other users too.

shashank1300:
# 🦟 Bug fix

Fixes #627

## Summary
Installs libGrid3D.so (a legacy alias for libGridConfig.so) by copying the already installed libGridConfig.so using install(CODE ...) instead of install(FILES ...). This ensures the build directory RUNPATH is correctly stripped, matching other installed plugins.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/plugins/grid_config/CMakeLists.txt
+++ b/src/plugins/grid_config/CMakeLists.txt
@@ -7,8 +7,14 @@ gz_gui_add_plugin(GridConfig
     gz-rendering${GZ_RENDERING_VER}::gz-rendering${GZ_RENDERING_VER}
 )
 
-# Also install as Grid3D, which was a legacy plugin with a subset of features
-install (
-  FILES $<TARGET_FILE:GridConfig>
-  RENAME ${CMAKE_SHARED_LIBRARY_PREFIX}Grid3D${CMAKE_SHARED_LIBRARY_SUFFIX}
-  DESTINATION ${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR})
+# Also install Grid3D (a legacy plugin with a subset of features) by copying the installed GridConfig library. This ensures the correct RUNPATH is used.
+
+install(CODE "
+  if(EXISTS \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGridConfig.so\")
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGridConfig.so\"
+    \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGrid3D.so\")
+  else()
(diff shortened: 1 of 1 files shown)


wentasah (approved):
This is clever! I missed the existence of `install(CODE`.

I confirm that this resolves my original issue. Thank you.

src/plugins/grid_config/CMakeLists.txt
```diff
@@ -7,8 +7,14 @@ gz_gui_add_plugin(GridConfig
...
+
+install(CODE "
+  if(EXISTS \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGridConfig.so\")
+    execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGridConfig.so\"
+    \"\${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGrid3D.so\")
+  else()
+    message(WARNING \"Source file for Grid3D (legacy plugin) copy not found, skipping copy: \${CMAKE_INSTALL_PREFIX}/${GZ_GUI_PLUGIN_RELATIVE_INSTALL_DIR}/libGridConfig.so\")
+  endif()
+")
```
arjo129, line 20: Knit: New line at end of file.

arjo129 (changes requested):
Minor knit.

This looks correct to me and works for me. However it'd be great if @j-rivero or someone with more knowledge of cmake and our release systems takes a look to ensure we aren't breaking anything.

src/plugins/grid_config/CMakeLists.txt
shashank1300: Done. Thanks for the review.

PR #782, opened 2026-09-08
Fix MinimalScene render synchronization during shutdown

tonynajjar:
# 🦟 Bug fix

## Summary

The GUI can hang on shutdown because queued rendering callbacks can overwrite `RenderSync::ShuttingDown`. This leaves the render thread waiting for a Qt handshake that will never happen, while the GUI waits for the thread to exit.

This fix keeps `ShuttingDown` terminal and wakes all synchronization waiters during shutdown. No public API changes.

### Validation

- A local C++ harness reproduced both worker-side and Qt-side hangs before the fix. Both pass afterward, along with the normal rendering handshake.
- `INTEGRATION_minimal_scene` and local `codecheck` passed. Cppcheck reported informational missing-system-header messages.
- Tested the fix on gz-gui 10.0.0 with gz-sim 10.5.0 / ROS 2 Lyrical: the GUI exited within two seconds of SIGINT in both runs.

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty

The same affected code exists in `gz-gui10`, where this fix was also tested. Older versions have not been evaluated.

## Checklist

- [ ] Added a screen capture or video (as needed)
- [ ] Added tests (local harness only; no committed regression test)
- [ ] Updated documentation (not needed)
- [ ] Updated migration guide (not needed)
- [ ] Consider updating Python bindings (not applicable)
- [x] `codecheck` passed locally
- [ ] All tests passed (only targeted tests run locally)
- [ ] Updated Bazel files (not applicable; no new files)
- [ ] Help review another open pull request
- [x] Was GenAI used to generate this PR? Yes

Assisted-by: GitHub Copilot

**Note to maintainers**: Use **Squash-Merge**, retaining `Signed-off-by`, `Assisted-by`, and any `Generated-by` trailers.

**Backports:** Use **Rebase and Merge**.

### DIFF
--- a/src/plugins/minimal_scene/MinimalScene.cc
+++ b/src/plugins/minimal_scene/MinimalScene.cc
@@ -249,13 +249,17 @@ void RenderSync::WaitForQtThreadAndBlock(std::unique_lock<std::mutex> &_lock)
   { return this->renderStallState == RenderStallState::WorkerCanProceed ||
            this->renderStallState == RenderStallState::ShuttingDown; });
 
+  if (this->renderStallState == RenderStallState::ShuttingDown)
+    return;
+
   this->renderStallState = RenderStallState::WorkerIsProceeding;
 }
 
 /////////////////////////////////////////////////
 void RenderSync::ReleaseQtThreadFromBlock(std::unique_lock<std::mutex> &_lock)
 {
-  this->renderStallState = RenderStallState::QtCanProceed;
+  if (this->renderStallState != RenderStallState::ShuttingDown)
+    this->renderStallState = RenderStallState::QtCanProceed;
   _lock.unlock();
   this->cv.notify_one();
 }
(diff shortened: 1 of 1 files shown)


azeey:
Thanks for the contribution! Can you please fill out the PR template?

tonynajjar:
yes done!

azeey:
Did you have a reliable way to reproduce the GUI hanging during shutdown?

tonynajjar:
Unfortunately no "reliable" way and i couldn't reproduce it simply with the basic "shapes" gazebo example.

On our simulation it reproduces about 50% of the time. I did not try to make a minimal reproducible example to understand which exact part of our simulation causes the GUI to hang.
I can try to do that with some AI help if it's important to you

azeey:
It would be great if you can since I can't seem to reproduce it myself.

tonynajjar:
@azeey could you try with `taskset -c 0 gz sim -v 4 -r shapes.sdf`

This pins the process to one cpu core to simulate realistic load and seems to help reproduce more often. You might need to try 5-10 times

